PR #8, opened 2015-02-04
Include ISIS Security extensions to Shiro for RBAC and SSO

lucamilanesio:
The Apache ISIS project made a very interesting extension to Shiro
allowing the automatic fetch of LDAP Groups and assigning them 
to roles in order to resolve the user’s permissions.

This extension makes sense _inside_ Shiro as it can be reused in many
other context even outside ISIS.

Additionally I have extended the LDAP Realm and Authentication Filter
in order to allow SSO with an external authentication system resolved
through an HTTP Proxy (e.g. Kerberos, but it could be anything else, 
including commercial products such as Site-Minder) so that they can
be used in conjunction.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,42 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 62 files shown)


bdemers:
The patch is: 
https://github.com/apache/shiro/pull/8/commits/91b5833a2ae88eafffc95ca33a264650eb053ec5

it doesn't render correctly after the switch from svn to git.

fpapon:
@bdemers do you think this PR can be merged from the patch? May be we could create a new PR from the patch. Thoughts?

asfgit:
https://builds.apache.org/job/Shiro-pr/55/

bdemers:
@fpapon I don't think this change is generic enough to live in the Shiro repo.  There is probably some thing we could do on the Shiro side.

I'm not sure the Apache + GitHub integration lets us close issues either (other than opening an infra request or using a `Fixes: #8` trailer)

bdemers:
Closing due to inactivity, please comment if you feel it should be reopened

PR #12, opened 2016-03-26
Implement SaltStyle.CRYPT

manticore-projects:
Implement SaltStyle.CRYPT fro the JdbcRealm.
Add some basic tests.


core/src/main/java/org/apache/shiro/realm/jdbc/JdbcRealm.java
```diff
@@ -227,9 +241,72 @@ protected AuthenticationInfo doGetAuthenticationInfo(AuthenticationToken token)
...
+                                else if (hashAlgorithm.equals("5")) 
+                                    hashedCredentialsMatcher.setHashAlgorithmName(Sha256Hash.ALGORITHM_NAME);
+                                else if (hashAlgorithm.equals("1")) 
+                                    hashedCredentialsMatcher.setHashAlgorithmName(Md5Hash.ALGORITHM_NAME);
+                                else if (hashAlgorithm.equals("2")) 
+                                    throw new AuthenticationException("Requested 'Blowfish' algorithm is not supported. Can not validate the token.");
+                                else if (hashAlgorithm.equals("2a")) 
+                                    throw new AuthenticationException("Requested 'eksblowfish' algorithm is not supported. Can not validate the token.");
+
+                                setCredentialsMatcher(credentialsMatcher);
```
bdemers: This is not thread safe.
Setting the `credentialsMatcher` here could have problems if the table contains multiple types of algorithms.

bmarwell:
Digging up this old PR, I recently implemented a shiro2 crypto format, which is based on the unix formats and uses the same identifiers. Maybe we can use those here, too? @bdemers ?

bdemers:
@bmarwell that sounds like a great idea!

Also, @manticore-projects, my previous review message seems a bit harsh when looking back at it.
Sorry about that! I should have started with a Thank you for sending us that PR!

manticore-projects:
> Also, @manticore-projects, my previous review message seems a bit harsh when looking back at it.
> Sorry about that! I should have started with a Thank you for sending us that PR!

No big deal and all good. If is is not threadsafe, we need to deal with it and when you point me on the line of code or illustrate your concern, I would work on it. (I know what threadsafe means, but I did not see how two threads would interfere here.)

core/src/main/java/org/apache/shiro/realm/jdbc/JdbcRealm.java
manticore-projects: Thanks, I get it now and I should have thought of that before:  the method ` AuthenticationInfo doGetAuthenticationInfo(AuthenticationToken token)` is called many times from various users and we should not create any Objects here, but only when the Realm is set up. (Sorry, this took me a while. I have had forgotten that I wrote that stuff 5 years ago).

I think, it would be save to create a Set of possible CredentialMatchers (based on the algorithms) when the Realm is created and then just use from this Set. Will look into it.

bmarwell: > I think, it would be save to create a Set of possible CredentialMatchers (based on the algorithms) when the Realm is created and then just use from this Set. Will look into it.

Please don't create static Sets, they require manual maintanance. The class `Shiro2CryptFormat` will already do what you need:
https://github.com/apache/shiro/blob/1c97815737fee587729b1d3827814f3f58d2b6ec/crypto/hash/src/main/java/org/apache/shiro/crypto/hash/format/Shiro2CryptFormat.java#L114-L135

I think we should extract the last few lines into it's own method, and we have a suitable method you can use:
https://github.com/apache/shiro/blob/1c97815737fee587729b1d3827814f3f58d2b6ec/crypto/hash/src/main/java/org/apache/shiro/crypto/hash/format/Shiro2CryptFormat.java#L129-L134

WDYT?

bmarwell: Btw, this class did not exist when you wrote this PR. I created it a few weeks ago for Shiro 2 explicitly. ;-)

bmarwell:
Ok, how about: Let's open an issue over at https://issues.apache.org/jira/browse/SHIRO, describe the idea behind this PR in general, rebase it, use the new crypto classes, and get it into 2.0.x? :)

PR #17, opened 2016-05-29
SHIRO-300 Add protected setParts(list) method

mattbishop:
WildcardPermission subclasses can create their own parts/subparts list and pass them in with this method.

See https://issues.apache.org/jira/browse/SHIRO-300

### DIFF
--- a/core/src/main/java/org/apache/shiro/authz/permission/WildcardPermission.java
+++ b/core/src/main/java/org/apache/shiro/authz/permission/WildcardPermission.java
@@ -188,6 +188,10 @@ public class WildcardPermission implements Permission, Serializable {
         return this.parts;
     }
 
+    protected void setParts(List<Set<String>> parts) {
+        this.parts = parts;
+    }
+
     /*--------------------------------------------
     |               M E T H O D S               |
     ============================================*/


bdemers:
Easy enough PR. 👍 

Someone else mentioning needing something like this too, basically to avoid re-parsing strings.
Do you have a use-case or child class you can share (doesn't need to be in this PR, I'm just interested)?

mattbishop:
Here is my nasty hack I am using until this PR comes in. Note the use of Field.set()!

```
package com.example.authorization.permission;

import java.lang.reflect.Field;
import java.util.List;
import java.util.Set;

import com.google.common.collect.ImmutableList;

import org.apache.commons.lang3.reflect.FieldUtils;
import org.apache.shiro.authz.permission.WildcardPermission;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

/**
	Stores a precomputed parts list. This class will no longer be needed once
	<a href="https://issues.apache.org/jira/browse/SHIRO-300">this feature is implemented in Shiro.</a>.
 */
public class PartsListWildcardPermission extends WildcardPermission {

	private static final Logger LOG = LoggerFactory.getLogger(PartsListWildcardPermission.class);

	private static final Field PARTS_FIELD = FieldUtils.getField(WildcardPermission.class, "parts", true);


	/**
	 * Default constructor.
	 */
	public PartsListWildcardPermission() {
		//nothing to do
	}

	/**
	 * Constructor. Takes precomputed part sets and uses them for wildcard permissions.
	 * @param parts permission parts
	 */
	public PartsListWildcardPermission(final List<Set<String>> parts) {
		this();
		setParts(parts);
	}

	/**
	 * Set the parts list.
	 * @param parts the permission parts.
	 */
	public void setParts(final List<Set<String>> parts) {
		ImmutableList<Set<String>> copy = ImmutableList.copyOf(parts);
		/*
			Shiro's WildcardPermission does a lot of direct-access use of parts internally, so we have to write the field as it is
			private and without a setter.
		 */
		try {
			PARTS_FIELD.set(this, copy);
		} catch (IllegalAccessException e) {
			LOG.error("Cannot store the parts field", e);
		}
	}
}```

PR #18, opened 2016-05-29
SHIRO-564 more efficient case-insensitive wildcard strings

mattbishop:
WildcartPermission will create the subparts list twice when it should consider permissions case-insensitively.

https://issues.apache.org/jira/browse/SHIRO-564


bdemers:
👍

PR #20, opened 2016-06-14
activeDirectoryRealm.principalSuffix isn't honoured

prabhjyotsingh:
While using shiro for auth in my project I noticed that, activeDirectoryRealm.principalSuffix is not honoured for authentication, however, for getRoleNamesForUser it does append the same.

Looking at the tcpdump, it appears activeDirectoryRealm.principalSuffix isn't honoured. As a result, the bind request searches for user_name, rather than user_name@company.com. Also note that if I try to login as username "user_name@company.com", I do authenticate as I'm found in the directory server

### DIFF
--- a/core/src/main/java/org/apache/shiro/realm/activedirectory/ActiveDirectoryRealm.java
+++ b/core/src/main/java/org/apache/shiro/realm/activedirectory/ActiveDirectoryRealm.java
@@ -105,7 +105,12 @@ public class ActiveDirectoryRealm extends AbstractLdapRealm {
         // Binds using the username and password provided by the user.
         LdapContext ctx = null;
         try {
-            ctx = ldapContextFactory.getLdapContext(upToken.getUsername(), String.valueOf(upToken.getPassword()));
+            String userPrincipalName = upToken.getUsername();
+            if (this.principalSuffix != null) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/shiro/realm/activedirectory/ActiveDirectoryRealm.java
```diff
@@ -105,7 +105,12 @@ protected AuthenticationInfo queryForAuthenticationInfo(AuthenticationToken toke
         // Binds using the username and password provided by the user.
         LdapContext ctx = null;
         try {
-            ctx = ldapContextFactory.getLdapContext(upToken.getUsername(), String.valueOf(upToken.getPassword()));
+            String userPrincipalName = upToken.getUsername();
+            if (this.principalSuffix != null) {
+                userPrincipalName = upToken.getUsername() + this.principalSuffix;
+            }
```
bdemers: This _should_ be already happening here: https://github.com/apache/shiro/blob/master/core/src/main/java/org/apache/shiro/realm/ldap/DefaultLdapContextFactory.java#L221-L222

Is this not happening in your setup ?

prabhjyotsingh:
I see this working now, not sure why it was not working earlier. Closing this.

PR #23, opened 2016-07-04
SHIRO-570: Only accept a cookie value when the request uses the proper path

ankon:


web/src/main/java/org/apache/shiro/web/servlet/SimpleCookie.java
```diff
@@ -362,8 +362,14 @@ public String readValue(HttpServletRequest request, HttpServletResponse ignored)
         String value = null;
         javax.servlet.http.Cookie cookie = getCookie(request, name);
         if (cookie != null) {
-            value = cookie.getValue();
-            log.debug("Found '{}' cookie value [{}]", name, value);
+            // Validate that the cookie is used at the correct place.
+            String path = StringUtils.clean(getPath());
+            if (path != null && !request.getRequestURI().startsWith(path)) {
```
bdemers: Should we just check if `getPath()` != `calcualtePath(...)` ?

ankon: I might be missing something, but if `getPath()` returns a non-null/non-empty string, then `calculatePath(request)` would return that value, which would trivially be equal?

ankon:
While looking at the question about `#calculatePath()`, I checked the rules for path matching for browsers, and found an edge case: the logic would accept a prefix that is not a complete path, for example it would accept requests from '/foobar/something.jsp' as matching for the cookie with a path of '/foo'.

bdemers:
+1 🍪 !

PR #24, opened 2016-07-06
WIP: CDI feature import SHIRO-337

bdemers:
I spent about a day looking into this, the original code was pretty good, but I ran into a few issues when adding a sample project 

Added CDI [README.md](support/cdi/README.md)
Added TODO's:
- [support/cdi](support/cdi/TODO.md)
- [samples](samples/cdi/TODO.md)

Comments above are just initial thoughts and are here to start discussion.

@hwellmann, I know it has been a while since you looked into this, but your thoughts would be appreciated


bdemers:
**NOTE:** history is in tacked at the moment, but will be squashed before merging

panchitoboy:
Hi, 

Did you have any estimations for this merge?

I have a library (shiro-jwt) to use apache shiro with Json Web Tokens. I use pax-shiro-cdi-web but i want to remove the dependency.

Also if you are interested i can help to integrate my library in the project. I implement a JWTOrFormAuthenticationFilter that could be usefull as default filter in REST applications.

Regards,

bdemers:
I was going to take another look at it in the next couple of weeks.  I've been looking into updating spring and guice support as well.  I'm hoping those will dovetail a bit with this, but priority wise, I want to focus on getting current modules updated first.

That said, if you want to take a shot at this, we would love the help.  Same with adding JWT support.  Send over a pull request and we can start that thread.

dmaidaniuk:
Hi guys, 

We also used in our application Shiro with CDI a lot. Thus I would like to help with given task. We looked into `pax-shiro-cdi-web` and `pax-shiro-cdi` and have ideas how to make them even more CDI-oriented, without need for `shiro-web` dependency, e.g. own `Environment` implementation etc. I guess I will start with pulling of existing changes and see how to add our improvements there.

Regards,
Dmytro

gilbertoca:
@bdemers @dmaidaniuk Any hope to include this in the final 1.4.0 release?

bdemers:
@gilbertoca  I poked around with this a few weeks ago, and I thought I had updated this thread
Take a look at this branch: https://github.com/apache/shiro/tree/cdi-idea (I just rebased it to master)

It was the result of a few late nights of hacking around.  I was originally trying to see if there some base code we could reuse for Spring, Guice, and CDI. But there didn't seem to be (unless we want to annotate everything that is).

Anyway, I'd like to hear thoughts on this approach vs requiring a `shiro.ini`.

gilbertoca:
My background is of someone who maintains webapps using shiro (And possibly will not contribute much). 
Initially we make the app work with a simple authe/autho setup. After we need to make some customization, mainly in the realm area. This customization is intensively done in the shiro,ini. Specifically in JEE containers, that customization needs to be CDI aware and inevitably we will start with a shiro.ini. 
Looking at the web sample I could see you did use shiro.ini - so this initial CDI support needs/requires one. 
I'm sorry I can't be of more help to you in this regard.

bdemers:
@gilbertoca Thanks, that is good feedback.

After looking at it again, it looks like the `cdi-web` sample wasn't really updated to show off anything CDI, I'll fix that soon.  But take a look at [other example](https://github.com/apache/shiro/blob/cdi-idea/samples/cdi-core/src/main/java/org/apache/shiro/samples/cdi/RealmProducer.java) and see if that would fit your needs.

This example still uses an INI file, but is ONLY used for the IniRealm's users (I'll probably update that too, to avoid any confusion)

dmaidaniuk:
Hi @bdemers and @gilbertoca,

Sorry for so big delay. I saw last changes in `cdi-idea` branch and they differ from original PAX implementation a bit. I will see more closely to them in nearest days. CDI applications can communicate with clients not only through HTTP requests, but by other channels, e.g. REST or EJB remote calls. Thus also required to test such scenarios also, when only `cdi-core` support library used.
Also would be good to reorganize common dependencies and move them to cdi-parent POM, e.g. `cdi-api` and `javax.annotation-api`.

Regards,
Dmytro

bdemers:
Thanks for the feedback @dmaidaniuk !

lprimak:
Superseded by Jakarta EE (cdi, ee) modules

PR #25, opened 2016-07-06
WIP: Ability to set a property value with an enum in shiro.ini file

bdemers:
SHIRO-425 (applied patch to master)
This enum support could be limited to property only, but that might cause confusion.

Currently supported via this patch:

``` ini
[main]
myObject = com.foo.bar.Type
myObject.enumProperty = ENUM_VALUE1
```

This is really cool, but as a user I would also expect the following to work:

``` ini
[main]
myObject = com.foo.bar.Type
myObject.enumListProperty = ENUM_VALUE1, ENUM_VALUE2
```

``` ini
[main]
myObject = com.foo.bar.Type
myObject.enumMapProperty = one:ENUM_VALUE1, two:ENUM_VALUE2
```

Granted the Map example might be a bit contrived, but less so for the list.

I've added a test for this.

Thoughts ?

### DIFF
--- a/config/ogdl/src/main/java/org/apache/shiro/config/ReflectionBuilder.java
+++ b/config/ogdl/src/main/java/org/apache/shiro/config/ReflectionBuilder.java
@@ -422,6 +422,17 @@ public class ReflectionBuilder {
             throw new ConfigurationException(msg, e);
         }
     }
+    //@since 1.3 : be able to set an enum value for an object property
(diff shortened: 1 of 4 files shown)


bdemers:
cc: @leleuj

leleuj:
This looks good to me

fpapon:
supercedes by https://github.com/apache/shiro/pull/199
We can close this one

PR #28, opened 2016-07-07
WIP: Applying patch from SHIRO-330  Encrypted Cookie

bdemers:
Creating pull request, because it is easier then dealing with patch files.

### DIFF
--- a/web/src/main/java/org/apache/shiro/web/servlet/SimpleCookie.java
+++ b/web/src/main/java/org/apache/shiro/web/servlet/SimpleCookie.java
@@ -215,7 +215,7 @@ public class SimpleCookie implements Cookie {
         addCookieHeader(response, name, value, comment, domain, path, maxAge, version, secure, httpOnly);
     }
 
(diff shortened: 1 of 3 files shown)


web/src/main/java/org/apache/shiro/web/servlet/SimpleCookie.java
```diff
@@ -215,7 +215,7 @@ public void saveTo(HttpServletRequest request, HttpServletResponse response) {
         addCookieHeader(response, name, value, comment, domain, path, maxAge, version, secure, httpOnly);
     }
 
-    private void addCookieHeader(HttpServletResponse response, String name, String value, String comment,
+    protected void addCookieHeader(HttpServletResponse response, String name, String value, String comment,
```
bdemers: Instead of changing this method signature, the same thing could probably be done buy overriding `buildHeaderValue()`

bdemers:
See: https://issues.apache.org/jira/browse/SHIRO-330

I think if Shiro takes on supporting encrypted cookies, we need to support signing them.

PR #29, opened 2016-07-11
Hacking: EventBus in Guice QuickStart

bdemers:
This is NOT where this code will live.  This is the result of a little hacking to see how one would register event listeners in a Guice environment.

### DIFF
--- a/samples/quickstart-guice/src/main/java/QuickstartShiroModule.java
+++ b/samples/quickstart-guice/src/main/java/QuickstartShiroModule.java
@@ -16,13 +16,64 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+import com.google.inject.Key;
+import com.google.inject.Provider;
 import com.google.inject.Provides;
+import com.google.inject.TypeLiteral;
+import com.google.inject.matcher.Matchers;
+import com.google.inject.multibindings.Multibinder;
+import com.google.inject.name.Names;
+import com.google.inject.spi.InjectionListener;
+import com.google.inject.spi.TypeEncounter;
+import com.google.inject.spi.TypeListener;
 import org.apache.shiro.config.Ini;
+import org.apache.shiro.config.event.LoggingBeanEventListener;
+import org.apache.shiro.event.Event;
+import org.apache.shiro.event.EventBus;
+import org.apache.shiro.event.Subscribe;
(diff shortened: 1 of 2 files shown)


samples/quickstart-guice/src/main/java/QuickstartShiroModule.java
```diff
@@ -16,22 +16,94 @@
...
+import java.util.Set;
 
 public class QuickstartShiroModule extends ShiroModule {
+
+    private EventBus eventBus;
+
+    @Provides
+    @Singleton
+    public EventBus getEventBus(@Named("eventListeners") Set<Object> eventListeners) {
+        // ignore the eventListeners, we just want to force the initial loading of these.
```
bdemers: This smells... 

I'm not sure how to force the eventListeners to be loaded, this works, but is ugly.

samples/quickstart-guice/src/main/java/QuickstartShiroModule.java
```diff
@@ -16,22 +16,94 @@
...
+
     protected void configureShiro() {
+
+        bindConstant().annotatedWith(Names.named("shiro.securityManager.eventBus")).to(EventBus.class);
+        expose(EventBus.class);
+
+        configureEventListeners(Multibinder.newSetBinder(binder(), Object.class, Names.named("eventListeners")));
+
+
+        bind(Key.get(LoggingBeanEventListener.class));
```
bdemers: this listener is referenced twice, once here, and once in the `configureEventListeners` method.

samples/quickstart-guice/src/main/java/QuickstartShiroModule.java
```diff
@@ -16,22 +16,94 @@
...
-    @Provides
-    Ini loadShiroIni() {
-        return Ini.fromResourcePath("classpath:shiro.ini");
+    protected void configureEventListeners(Multibinder<Object> eventListenerMultibinder ) {
+
+        eventListenerMultibinder.addBinding().to(Key.get(LoggingBeanEventListener.class));
     }
+
+
+    private class SubscribedEventTypeListener implements TypeListener {
```
bdemers: missing javadoc... but...
This TypeListener checks for the existence of methods with `@Subscribe`, and then calls EventBus.register(), after they have been created.

bdemers:
cleaned up and already merged into 1.4

PR #30, opened 2016-07-11
WIP: First pass at using the new EventBus to fire AuthC events.

bdemers:
I'm digging the simplicity of the `EventBus`, but currently it is only `really`\* used by the `ReflectionBuilder`

This is attempt to see what publishing auth events from the `AuthenticationRealm` would look like.
- The plumbing exists for custom realms, to make full use of the EventBus, it is just that none of the stock `Realms` make direct use of it at the moment.

### DIFF
--- a/core/src/main/java/org/apache/shiro/realm/AuthenticatingRealm.java
+++ b/core/src/main/java/org/apache/shiro/realm/AuthenticatingRealm.java
@@ -28,6 +28,7 @@ import org.apache.shiro.authc.credential.CredentialsMatcher;
 import org.apache.shiro.authc.credential.SimpleCredentialsMatcher;
 import org.apache.shiro.cache.Cache;
 import org.apache.shiro.cache.CacheManager;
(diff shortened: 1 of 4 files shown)


fpapon:
@bdemers I like this one, we can add it to the 2.0.0 work.
Thoughts?

bdemers:
@fpapon yeah!
Does the `EventBusAware` and the other `*Aware` casting usage of Shiro work well with OSGi?

fpapon:
> @fpapon yeah!
> Does the `EventBusAware` and the other `*Aware` casting usage of Shiro work well with OSGi?

I never try but we have an EventAdmin in OSGi and I think it would be nice to use it. I will take a look ;)

bmhm:
Great! I am really looking forward to seeing other events, such as "configurationUpdatedEvent" (if that ever happened) or events about the realms from a SecurityManager perspective, as those could be useful for M(X)Beans.

fpapon:
I'm also thinking about to sent events to external bus like Kafka.

PR #31, opened 2016-07-13
SHIRO-361 Disable encoding JSESSIONID in request URL by default.

bdemers:
This can be turned on by setting: `sessionManager.sessionIdUrlRewritingEnabled` to false
in the shiro.ini (or equivalent configuration)

This is a stop gap on the way to SHIRO-360


bdemers:
`stormpathRealm.authenticationCachingEnabled = true` now defaults to true (no change in behavior)

merged to 1.3.x and master

PR #32, opened 2016-07-14
WIP: Attempt at getting guice 3 and 4 both supported

bdemers:
This still needs some cleaning up, but I was hacking on this last night, and I _think_ this will work.
It needs a few eyes, and I have a couple questions, that I'll annotate here.

This does not have separate test module like #26 (to enable guice4 you can just use the profile `guice4`, I don't like this, but it works for hacking around). We should add a test module like #26, or something similar.

### DIFF
--- a/samples/guice/src/main/java/org/apache/shiro/samples/guice/SampleShiroServletModule.java
+++ b/samples/guice/src/main/java/org/apache/shiro/samples/guice/SampleShiroServletModule.java
@@ -31,6 +31,7 @@ import org.apache.shiro.web.mgt.DefaultWebSecurityManager;
 import org.apache.shiro.web.mgt.WebSecurityManager;
 
 import javax.inject.Singleton;
+import javax.servlet.Filter;
 import javax.servlet.ServletContext;
 import java.net.MalformedURLException;
 import java.net.URL;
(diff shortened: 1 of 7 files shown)


support/guice/src/main/java/org/apache/shiro/guice/ShiroModule.java
```diff
@@ -52,7 +52,7 @@
 
 	private Set<Destroyable> destroyables = Collections.newSetFromMap(new WeakHashMap<Destroyable, Boolean>());
     public void configure() {
-        this.binder().requireExplicitBindings();
+//        this.binder().requireExplicitBindings();
```
bdemers: @jbunting I'm not sure why this was needed, but seems to be causing problems with guice4 in the ShiroModuleTests (and the tests pass without this with both guice 3 and 4)

support/guice/src/main/java/org/apache/shiro/guice/web/ShiroWebModule.java
```diff
@@ -228,24 +323,52 @@ protected void bindWebEnvironment(AnnotatedBindingBuilder<? super WebEnvironment
      * @param keys
      */
     @SuppressWarnings({"UnusedDeclaration"})
+    @Deprecated
     protected final void addFilterChain(String pattern, Key<? extends Filter>... keys) {
-        filterChains.put(pattern, keys);
+
```
bdemers: This would still need tests

samples/guice/src/main/java/org/apache/shiro/samples/guice/SampleShiroServletModule.java
```diff
@@ -56,7 +57,8 @@ protected void configureShiroWeb() {
         this.addFilterChain("/login.jsp", AUTHC);
         this.addFilterChain("/logout", LOGOUT);
         this.addFilterChain("/account/**", AUTHC);
-        this.addFilterChain("/remoting/**", AUTHC, config(ROLES, "b2bClient"), config(PERMS, "remote:invoke:lan,wan"));
+
+        this.addFilterChain("/remoting/**", filterConfig(AUTHC), filterConfig(ROLES, "b2bClient"), filterConfig(PERMS, "remote:invoke:lan,wan"));
```
bdemers: `filterConfig` replaces `config`, `config` will only work with guice3

bdemers:
cleaned up and already merged into 1.4

PR #38, opened 2016-09-24
SHIRO-586 - Can't Search For Groups In Active Directory Without A System User

opticyclic:
Change getRoleNamesForUser from private to protected so that it can be used in sub-classes.
This will enable implementing a custom Realm (that sub-classes ActiveDirectoryRealm) able to access groups/roles without needing a system user.


bdemers:
@opticyclic can you close this, it did not auto-close after merging

PR #41, opened 2016-10-07
When using more than 1 realm for authentication ModularRealmAuthenticator would log a stack trace each time a user enters a wrong password. Turned down the logging level for this to happen from WARN to TRACE.

scsynergy:
When using more than 1 realm for authentication ModularRealmAuthenticator would log a stack trace each time a user enters a wrong password. Turned down the logging level for this to happen from WARN to TRACE.


bdemers:
:+1:
Thanks!

bdemers:
I found an older issue in JIRA (linked in SHIRO-592), basically this was discussed before, though the comments in SHIRO-462 are valid, that could also be resolved by increasing the log level (as opposed to decreasing it)

PR #43, opened 2016-10-17
Update features.xml

guilherfp:
Fix for Karaf 4.0.5

There is not a version <4 of the war bundle at karaf repositories.

### DIFF
--- a/support/features/src/main/resources/features.xml
+++ b/support/features/src/main/resources/features.xml
@@ -27,7 +27,7 @@
 
     <feature name="shiro-web" version="${project.version}" resolver="(obr)">
         <feature version="${project.version}">shiro-core</feature>
-        <feature version="[2,4)">war</feature>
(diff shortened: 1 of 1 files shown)


bdemers:
@guilherfp thanks!

Is there an easy way we can test the features file (and bundle configs).  These pull requests are a great help, but I'd like to see how we can better test for these things.

bdemers:
@guilherfp ^

gnodet:
I see that shiro-core requires the {{org.apache.commons.configuration2.interpol}} package.  Should it be flagged as optional ?

gnodet:
#63 supercedes this PR and in addition to fixing the features, it also provides build time validation.

fpapon:
@guilherfp thanks for the PR!

It's include in:

https://github.com/apache/shiro/pull/88

I close this one.

PR #45, opened 2016-10-29
Added private salt options to the Hasher Tool to allow users to specific

HipDragon:
This is for issue SHIRO-596, which i created to track this updated feature for the shiro-tools-hasher.

### DIFF
--- a/crypto/hash/src/main/java/org/apache/shiro/crypto/hash/DefaultHashService.java
+++ b/crypto/hash/src/main/java/org/apache/shiro/crypto/hash/DefaultHashService.java
@@ -21,6 +21,7 @@ package org.apache.shiro.crypto.hash;
 import org.apache.shiro.crypto.RandomNumberGenerator;
 import org.apache.shiro.crypto.SecureRandomNumberGenerator;
 import org.apache.shiro.util.ByteSource;
+import org.apache.shiro.util.ByteSource.Util;
 
 /**
  * Default implementation of the {@link HashService} interface, supporting a customizable hash algorithm name,
@@ -158,7 +159,7 @@ public class DefaultHashService implements ConfigurableHashService {
 
         ByteSource publicSalt = getPublicSalt(request);
         ByteSource privateSalt = getPrivateSalt();
-        ByteSource salt = ByteSource.Util.combine(privateSalt, publicSalt);
+        ByteSource salt = combine(privateSalt, publicSalt);
 
         Hash computed = new SimpleHash(algorithmName, source, salt, iterations);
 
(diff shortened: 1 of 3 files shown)


bdemers:
Cool! can you move the `combine` method to `ByteSource.Util` so both the `DefaultHashService` and `Hasher` can use the same method (or something similar)?

Thanks for cleaning up the deprecated `PosixParser` !

HipDragon:
I'll move the combine method to ByteSource.Util and update the Hasher and DefaultHashService to use it as well.

HipDragon:
This is ready to go now.

bdemers:
Great! I'll take another look tomorrow!

crypto/hash/src/main/java/org/apache/shiro/crypto/hash/DefaultHashService.java
```diff
@@ -233,45 +233,6 @@ protected ByteSource getPublicSalt(HashRequest request) {
...
-     * total salt during hash computation.  {@code privateSaltBytes} will be {@code null} }if no private salt has been
-     * configured.
-     *
-     * @param privateSalt the (possibly {@code null}) 'private' salt to combine with the specified extra bytes
-     * @param publicSalt  the extra bytes to use in addition to the given private salt.
-     * @return a combination of the specified private salt bytes and extra bytes that will be used as the total
-     *         salt during hash computation.
-     */
-    protected ByteSource combine(ByteSource privateSalt, ByteSource publicSalt) {
-
```
bdemers: One last thing, sorry it has taken so long to get to this.

This method signature needs remain the same for backwards compatibility, as well as the calling of this method on line `161`.

Just replace the body with a call to: `ByteSource.Util.combine()`

Thanks again!

HipDragon:
Made the changes and marked it as deprecated since 1.4 so that in the future it may be removed. It also allows people migrate away from it.

bdemers:
@HipDragon I feel like I'm dragging this out, and lead you down the wrong path.

I was going through this PR this morning, and I figured instead of duplicating the hashing logic or moving the combine logic to a Util method (which I admit, I suggested).  I used the `DefaultHashService` and a `SimpleHashRequest`

Take a look at #48, I squash your changes into this, and left your name on the commit.  I ran a quick manual test, and both #45 and #48 have the same output.  Let me know what you think of all of this, and sorry for sending you down the wrong path.

HipDragon:
I took a look at #48 and that solution was one i thought about as well. But didn't know if that was the right direction based on coding choices made in the code previously. Then being lead down the wrong rabbit hole i thought i made the right choice.

I'm just glad i could help get this feature included in the tool, and hope this gets merged into the next release. I'm hoping to find something else to improve and submit it as well in the future.

bdemers:
Sorry for the wait, my environment has been dirty for the past few days.

I've merged #48 

Please close this one (as it will not be closed automatically)

Thanks again @HipDragon !

PR #49, opened 2016-12-04
New AuthenticationStrategy - ShortCircuitFirstSuccessfulStrategy

ashokkumarta:
A new AuthenticationStrategy implementation - ShortCircuitFirstSuccessfulStrategy, to avoid looping through all the realms after a first successful authentication.

### DIFF
--- a/core/src/test/java/org/apache/shiro/authc/pam/ShortCircuitFirstSuccessfulStrategyTest.java
+++ b/core/src/test/java/org/apache/shiro/authc/pam/ShortCircuitFirstSuccessfulStrategyTest.java
@@ -1,0 +1,56 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.shiro.authc.pam;
+
+import static org.junit.Assert.assertNull;
+import org.junit.Before;
+import org.junit.Test;
+
+import org.apache.shiro.authc.AuthenticationException;
+import org.apache.shiro.authc.AuthenticationInfo;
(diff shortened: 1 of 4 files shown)


bdemers:
Nice!

There was a [similar thread](http://dev.shiro.apache.narkive.com/pZKBq1jV/making-the-modularrealmauthenticator-not-check-all-realms-for-firstsuccessfulstrategy) a few years ago on this. I'm guessing you have seen this already.

I think we should roll this into the, `FirstSuccessfulStrategy` disabled by default but enabled with a boolean.  That way we can make this the default operation for 2.0.

What to you think?

core/src/test/java/org/apache/shiro/authc/pam/ShortCircuitFirstSuccessfulStrategyTest.java
```diff
@@ -0,0 +1,56 @@
...
+import static org.junit.Assert.assertNull;
+import org.junit.Before;
+import org.junit.Test;
+
+import org.apache.shiro.authc.AuthenticationException;
+import org.apache.shiro.authc.AuthenticationInfo;
+import org.apache.shiro.authc.SimpleAuthenticationInfo;
+
+
+public class ShortCircuitFirstSuccessfulStrategyTest {
```
bdemers: Thank you for including a Test!!

ashokkumarta:
Yes. I had seen that thread of discussion and also responded to it [here](http://shiro-developer.582600.n2.nabble.com/Making-the-ModularRealmAuthenticator-not-check-all-realms-for-FirstSuccessfulStrategy-td7359001.html)

A new strategy looked better to me. That way we don't alter the way `FirstSuccessfulStrategy` worked, retaining full backward compatibility. And each strategy has a single definite behavior. 

But on second thought, what you suggest makes better sense. I don't see any use case where `FirstSuccessfulStrategy` in its current form adds value. Its behavior needs to be changed. As you had suggested, its better to build this behavior into `FirstSuccessfulStrategy` itslef, controlled through a boolean, rather then introducing a new strategy.  

Also was thinking if adding `FirstNSuccessfulStrategy` & `AtLeastNSuccessfulStrategy` to have min of 2, 3, ... n successful authentications would be beneficial. Any thoughts on this. 

Where can I read on the 2.0 roadmap?

ashokkumarta:
Created a pull request for boolean flag based break in `FirstSuccessfulAuthenticationStrategy`
[Pull request](https://github.com/apache/shiro/pull/50/files)

fpapon:
retest please

asfgit:
https://builds.apache.org/job/Shiro-pr/49/

fpapon:
@bdemers is this PR can be add to the 1.4.1?

bdemers:
@fpapon I'm a fan of #50 instead of this one, we can then roll that into the default (only) behavior for 2.0.
Thanks again @ashokkumarta!!

fpapon:
@bdemers ok I will check the orther PR ;)

bdemers:
supercedes by #184

PR #55, opened 2017-01-15
SslFilter with HTTP Strict Transport Security (HSTS)

raupachz:
Hi Brian,

we talked on the mailing list. For anyone not involved here is the summary.

HTTP Strict Transport Security (HSTS) would be a nice addition for all the SSL only sites out there. I think in recent years more and more pages have gone full SSL, with good reasons to do so. It is a bit problematic with SslFilter since this one is path based. If you go HSTS then everything on the site uses https. This might break thinks if you have a path with ssl and one without. You can do that with shiro but not with HSTS.

Some words on the implementation, I am not sure on some things.

I have never used EasyMock before. The test cases in SslFilterTest run fine, but maybe I did it very overcomplicated.

Is `postHandle` invoked even if `isAccessAllowed` returns false? The user agent has to ignore the HTTP header if it is send over HTTP, but I would prefer to send it only over HTTPS. So `isAccessAllowed` needs to return true. I could move the logic to isAccessAllowed, but it looked like the wrong place.

### DIFF
--- a/web/src/main/java/org/apache/shiro/web/filter/authz/SslFilter.java
+++ b/web/src/main/java/org/apache/shiro/web/filter/authz/SslFilter.java
@@ -112,11 +112,11 @@ public class SslFilter extends PortFilter {
      */
     @Override
     protected void postHandle(ServletRequest request, ServletResponse response)  {
-        if (hsts.enabled) {
+        if (hsts.isEnabled()) {
             StringBuilder directives = new StringBuilder(64)
                     .append("max-age=").append(hsts.getMaxAge());
             
-            if (hsts.includeSubDomains) {
+            if (hsts.isIncludeSubDomains()) {
                 directives.append("; includeSubDomains");
             }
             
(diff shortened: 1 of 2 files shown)


bdemers:
Yeah, you would need to use `preHandle()` Take a look at `AdviceFilter.doFilterInternal()`.

Though on this topic the filter actually executing is still based on the order filters are defined, and applied to paths.

``` ini
[urls]
/** = ssl
/foobar = ssl
/barfoo = authc
/abcxyz = authc, ssl
```

in the example above the path `/barfoo` does NOT execute the `ssl` filter, and `/abcxyz` executes the authc filter before the ssl one.  I had a similar problem when working on the Stormpath Shiro integration, and used a [ShiroPrioritizedFilterChainResolver](https://github.com/stormpath/stormpath-shiro/blob/master/extensions/servlet/src/main/java/com/stormpath/shiro/servlet/filter/ShiroPrioritizedFilterChainResolver.java) which is configured with a set of filters that will execute before any path filters.  So in this case, it would probably look something like:

``` ini
[main]
# enable hsts
ssl.hsts.enabled = true

# All paths go through ssl filter
priorityFilter.filters = ssl

[urls]
/** = authc
```

That said, this is not part of Shiro at the moment, and I think this pull request can be completed independent of this, but it may help both with your other question:

The issue with the path configuration of the ssl filter and path, does seem odd, but I think this solution makes it easier to configure and explain/configure.  I'd love to hear other thoughts on this though.

fpapon:
@bdemers I would like to add this PR to the next release.
I also like the ShiroPrioritizedFilterChainResolver from Stormpath that you mentioned. Any chance to add this in Shiro?

bdemers:
@fpapon 
Sure, I can add that!

fpapon:
@bdemers can I create a Jira for adding the `ShiroPrioritizedFilterChainResolver`?

bdemers:
Please do! sorry I lost track of this one.  I'm going to block out some time in the next week or so to focus on Shiro :)

fpapon:
> Please do! sorry I lost track of this one. I'm going to block out some time in the next week or so to focus on Shiro :)

done: https://issues.apache.org/jira/browse/SHIRO-779

fpapon:
@bdemers can you review please?

fpapon:
> +1
> We should probably add this to 1.6 too
> 
> We will need to add doc after we cut a release too, I'm assuming the `ini` config would look something like:
> 
> ```ini
> ssl.hsts.enabled = true
> ssl.hsts.includeSubDomains =true
> ```

Agree for the doc. 
I will add a 1.6 version on Jira and tag the issue to this version. We will just have to cherry-pick the commit.

bmhm:
retest this please

PR #56, opened 2017-01-24
[SHIRO-337] IDEA: CDI - no shiro.ini required

bdemers:
It was the result of a few late nights of hacking around. I was originally trying to see if there some base code we could reuse for Spring, Guice, and CDI. But there didn't seem to be (unless we want to annotate everything that is).

Fixes: #24


lprimak:
Superseeded by #695

PR #57, opened 2017-02-01
Add IpFilter for restricting access IP ranges

rswheeldon:
Add IpFilter for restricting access to resources from those coming from without (or outside) specific IP ranges.

Add IpAddressMatcher taken from Spring Security used for range tests

### DIFF
--- a/web/src/main/java/org/apache/shiro/web/filter/authz/IpAddressMatcher.java
+++ b/web/src/main/java/org/apache/shiro/web/filter/authz/IpAddressMatcher.java
@@ -26,7 +26,7 @@ import java.util.Arrays;
  * Both IPv6 and IPv4 addresses are supported, but a matcher which is configured with an
  * IPv4 address will never match a request which returns an IPv6 address, and vice-versa.
  *
- * @author Luke Taylor originally written for Spring Security
+ * @see <a href="https://github.com/spring-projects/spring-security/blob/master/web/src/main/java/org/springframework/security/web/util/matcher/IpAddressMatcher.java">Original Spring Security version</a>
  * @since 1.4.0
  */
 public final class IpAddressMatcher {
--- a/web/src/main/java/org/apache/shiro/web/filter/authz/IpAddressMatcher.java
+++ b/web/src/main/java/org/apache/shiro/web/filter/authz/IpAddressMatcher.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright 2002-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.shiro.web.filter.authz;
+
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.Arrays;
+
+/**
+ * Matches a request based on IP Address or subnet mask matching against the remote
+ * address.
+ * <p>
+ * Both IPv6 and IPv4 addresses are supported, but a matcher which is configured with an
+ * IPv4 address will never match a request which returns an IPv6 address, and vice-versa.
+ *
+ * @author Luke Taylor originally written for Spring Security
+ * @since 1.4.0
+ */
+public final class IpAddressMatcher {
+    private final int nMaskBits;
+    private final InetAddress requiredAddress;
+   
+    /**
+     * Takes a specific IP address or a range specified using the IP/Netmask (e.g.
+     * 192.168.1.0/24 or 202.24.0.0/14).
+     *
+     * @param ipAddress the address or range of addresses from which the request must
+     * come.
+     */
+    public IpAddressMatcher(String ipAddress) {
+        int i = ipAddress.indexOf('/');
+        if (i > 0) {
+            nMaskBits = Integer.parseInt(ipAddress.substring(i + 1));
+            ipAddress = ipAddress.substring(0, i);
+        } else {
+            nMaskBits = -1;
+        }
+        requiredAddress = parseAddress(ipAddress);
+    }
+
+    public boolean matches(String address) {
+        InetAddress remoteAddress = parseAddress(address);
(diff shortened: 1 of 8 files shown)


web/src/main/java/org/apache/shiro/web/filter/authz/IpAddressMatcher.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.Arrays;
+
+/**
+ * Matches a request based on IP Address or subnet mask matching against the remote
+ * address.
+ * <p>
+ * Both IPv6 and IPv4 addresses are supported, but a matcher which is configured with an
+ * IPv4 address will never match a request which returns an IPv6 address, and vice-versa.
+ *
+ * @author Luke Taylor originally written for Spring Security
```
bdemers: @Author tags should be removed, but a link to the original source would be great.

You can also update the [NOTICE](https://github.com/apache/shiro/blob/master/NOTICE) if needed. The current wording my be sufficient, but I'll leave this comment here, incase anyone else has thoughts on this.

web/src/main/java/org/apache/shiro/web/filter/authz/IpFilter.java
```diff
@@ -0,0 +1,142 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Collection;
+
+/**
+ * A Filter that requires the request to be from within a specific set of IP
+ * address ranges and / or not from with a specific (denied) set.
+ * <p/>
+ * Example config:
+ * <pre>
```
bdemers: Old Javadoc
The `[filters]` section is deprecated. but you can replace this with `[main]`

web/src/main/java/org/apache/shiro/web/filter/authz/IpFilter.java
```diff
@@ -0,0 +1,142 @@
...
+ * localLan.authorizedIps = 192.168.10.0/24
+ * localLan.deniedIps = 192.168.10.10/32
+ * <p/>
+ * [urls]
+ * /some/path/** = localLan
+ * # override for just this path:
+ * /another/path/** = localLan
+ * </pre>
+ *
+ * @since 1.0
```
bdemers: Should be 1.4

web/src/main/java/org/apache/shiro/web/filter/authz/IpSource.java
```diff
@@ -0,0 +1,39 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.shiro.web.filter.authz;
+
+import java.util.Collection;
+
+public interface IpSource {
```
bdemers: Add `@since` tag

web/src/test/java/org/apache/shiro/web/filter/authz/IpAddressMatcherTests.java
```diff
@@ -0,0 +1,78 @@
...
+ */
+package org.apache.shiro.web.filter.authz;
+
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
+
+import org.junit.Test;
+
+/**
+ * @author Modified by Richard Wheeldon from an original by Luke Taylor for Spring Security
```
bdemers: See comment about `@author` above

web/src/test/java/org/apache/shiro/web/filter/authz/IpAddressMatcherTests.java
```diff
@@ -0,0 +1,78 @@
...
+ * @since 1.4.0
+ */
+public class IpAddressMatcherTests {
```
bdemers: Thanks for including the tests as well!!

bdemers:
This should probably be added to [`DefaultFilter`](https://github.com/apache/shiro/blob/master/web/src/main/java/org/apache/shiro/web/filter/mgt/DefaultFilter.java) as well. Possibly using the name `ipRange` (not sure of that name though, `ips` seems too short, `ipFilter` doesn't match the current naming scheme, other ideas welcome)

bdemers:
@rswheeldon Thanks for working on this! I added a few minor things, nothing major.  You should also sign a [Apache CLA](http://apache.org/licenses/#clas).

rswheeldon:
CLA was signed years ago when I pushed a load of stuff to Apache FOP. It should still be valid. If not, I'll sign a new one.

bdemers:
Cool, I only took a quick look, I'll look again

web/src/main/java/org/apache/shiro/web/filter/authz/IpFilter.java
rswheeldon: Also changing in PortFilter - which is where I cribbed it from.

rswheeldon:
Updated with fixes for the above.

bdemers:
@rswheeldon 
I found the CLA, I must have fat fingered the search last time.

Any thoughts on the `DefaultFilter` naming ?

rswheeldon:
ip - it's only one character shorter than ssl. rest, port and user aren't exactly verbose either. If you agree, I'll add it.

bdemers:
`ip` works for me

rswheeldon:
Done. I don't really understand the Guice stuff though so I've done nothing more on that than make the unit tests pass.

rswheeldon:
What else is needed to get this into trunk? There are no changes I'm intending on making unless there's something that's not been addressed.

bdemers:
This looks good to me :+1:

col-panic:
Whats the status on this? Currently looking for this kind of functionality?!

bdemers:
@col-panic give this PR a shoot and let us know what you think!

mookkiah:
Hello @col-panic @bdemers - We are looking towards to have this IP filter option in our application design. But seeing this PR left open concerns me. Is there any reason or alternative.
If it is just the merge conflict, I can contribute by resolving it. Please let me know

fpapon:
@mookkiah feel free to resolve the conflict, we will merge it!

col-panic:
I switched my implementation to be behind an nginx that acts as reverse proxy. This allows me to separate this concerns from the filter configuration which is now done in nginx. Due to this I did not further spend time on it. sry.

mookkiah:
@fpapon Created new PR #219 after resolving merge conflict.
@col-panic I agree that we can block IP using infrastructure solution like nginx. That is an option in our design as well. Being an Apache Shiro user, I had expectation to have this solution (at least as a configurable option). 
Hope this is useful for someone who wants to use simple option.

fpapon:
@mookkiah thanks! I close this one.

PR #68, opened 2017-07-14
FormAuthenticationFilter capacity limit in same production environments

sysuzhang:
when our application was runing behind a nginx proxy. The request context was different with tomcat's context, for example my nginx proxy configuare is something like: 
 location ^~ /
        {
            proxy_pass http://MyTomcatServer/tomcatContext/;
        }
the session will lost because of cookie path change. though I know I can use proxy_cookie_path to change the cookie' path, I found it's still very useful when my class extend the FormAuthenticationFilter class. so I 
think it's betther to  provide two functions that are  very useful when your application running behind a proxy server which has  different context,  also very useful for subclasses pass parameters to the login page. 
How do you think?

### DIFF
--- a/web/src/main/java/org/apache/shiro/web/filter/AccessControlFilter.java
+++ b/web/src/main/java/org/apache/shiro/web/filter/AccessControlFilter.java
@@ -25,6 +25,7 @@ import org.apache.shiro.web.util.WebUtils;
 import javax.servlet.ServletRequest;
 import javax.servlet.ServletResponse;
 import java.io.IOException;
+import java.util.Map;
 
 /**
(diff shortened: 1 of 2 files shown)


bdemers:
Cool, we will take a closer look!

bdemers:
We lost track of this one, sorry about that.

I'm not sure if you are still watching this issue or not, but it looks like you could just cal
```java
WebUtils.issueRedirect(request, response, loginUrl, queryParams, contextRelative);
```
from your custom filter?  As the overloaded methods in this PR don't actually change the functionality.  Were you just trying to expose this to make it easier to call?

PR #73, opened 2017-10-25
shiro-guice depends on commons-beanutils

bentmann:
shiro-guice has a direct dependency on commons-beanutils (via `ShiroModule` -> `BeanTypeListener.beanUtilsBean`), would be nice to declare that fact properly.


bdemers:
Thanks @bentmann that is an easy one!  We will get it in for the next release.

bdemers:
Merged! Thanks again @bentmann !

PR #74, opened 2017-11-18
[SHIRO-640] queryForAuthenticationInfo(): resolve DN using user name

mephi42:
I am trying to use ActiveDirectoryRealm with searchFilter in order to make it possible to log in using e-mail (which is not part of DN). I see that this is partially supported in getRoleNamesForUser(), but not in queryForAuthenticationInfo().

This change make it fully work for me, but I have a feeling that it may disturb other users, so I'm willing to work on improving it.

### DIFF
--- a/core/src/main/java/org/apache/shiro/realm/activedirectory/ActiveDirectoryRealm.java
+++ b/core/src/main/java/org/apache/shiro/realm/activedirectory/ActiveDirectoryRealm.java
@@ -39,6 +39,7 @@ import javax.naming.directory.Attributes;
 import javax.naming.directory.SearchControls;
 import javax.naming.directory.SearchResult;
 import javax.naming.ldap.LdapContext;
+import java.rmi.Naming;
 import java.util.*;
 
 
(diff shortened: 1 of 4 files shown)


bdemers:
We may want to combine some of the functionality in the DefaultLdapRealm with the AD realm.  Take a look at: https://github.com/apache/shiro/blob/8acc82ab4775b3af546e3bbde928f299be62dc23/core/src/main/java/org/apache/shiro/realm/ldap/DefaultLdapRealm.java#L227

mephi42:
Why are DefaultLdapRealm and AD realm are separated altogether?
Wouldn't it make sense to merge them?

bdemers:
IIRC, ActiveDirectory typically uses one or two different schemas,
depending on how it is configured.  The schema for an LDAP server could be
just about anything (but typically follow a couple patterns).

We could make them share a lot more code through the AbstractLdapRealm

mephi42:
In this update I made two improvements:

* shared user DN substitution logic between AbstractLdapRealm and DefaultLdapRealm
* took care of potential compatibility issue, when callers take care of resolution themselves, by providing a new userSearchFilter property

bdemers:
Thanks @rubenvanwanzeele: #81

rubenvanwanzeele:
@bdemers What's the status of this? Is there any update on this functionality?

PR #79, opened 2018-02-26
Fix for SHIRO-621: REST filter bypassing matched path

ittiaminsite:
Use LinkedHashMap to maintain order of path to config map in filter chain in ShiroWebModule.

### DIFF
--- a/support/guice/pom.xml
+++ b/support/guice/pom.xml
@@ -29,6 +29,7 @@
     <modelVersion>4.0.0</modelVersion>
     <artifactId>shiro-guice</artifactId>
     <name>Apache Shiro :: Support :: Guice</name>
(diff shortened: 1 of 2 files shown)


support/guice/pom.xml
```diff
@@ -29,6 +29,7 @@
     <modelVersion>4.0.0</modelVersion>
     <artifactId>shiro-guice</artifactId>
     <name>Apache Shiro :: Support :: Guice</name>
+    <version>1.4.0-ittiaminsite</version>
```
bdemers: Looks good, but can you drop this line from the PR

Thanks!

bdemers:
Merged! Thanks!!

PR #86, opened 2018-09-13
replacement of "request.getRemoteHost()" by "WebUtils.getRemoteAddr()"

zhengyanbin:
help our‘s web server behind the proxy server get the true client ip

### DIFF
--- a/web/src/main/java/org/apache/shiro/web/util/WebUtils.java
+++ b/web/src/main/java/org/apache/shiro/web/util/WebUtils.java
@@ -92,6 +92,32 @@ public class WebUtils {
      */
     public static final String DEFAULT_CHARACTER_ENCODING = "ISO-8859-1";
 
+    public static String getRemoteAddr(ServletRequest servletRequest) {
+        HttpServletRequest request = toHttp(servletRequest);
+        if (request == null) {
+            return "unknown";
+        }
+        String ip = request.getHeader("x-forwarded-for");
+        if (ip == null || ip.length() == 0 || "unknown".equalsIgnoreCase(ip)) {
+            ip = request.getHeader("Proxy-Client-IP");
+        }
+        if (ip == null || ip.length() == 0 || "unknown".equalsIgnoreCase(ip)) {
+            ip = request.getHeader("X-Forwarded-For");
+        }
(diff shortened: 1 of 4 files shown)


web/src/main/java/org/apache/shiro/web/util/WebUtils.java
```diff
@@ -92,6 +92,32 @@
      */
     public static final String DEFAULT_CHARACTER_ENCODING = "ISO-8859-1";
 
+    public static String getRemoteAddr(ServletRequest servletRequest) {
+        HttpServletRequest request = toHttp(servletRequest);
+        if (request == null) {
+            return "unknown";
+        }
+        String ip = request.getHeader("x-forwarded-for");
```
bdemers: Why is `x-forwarded-for` used twice?

web/src/main/java/org/apache/shiro/web/util/WebUtils.java
```diff
@@ -92,6 +92,32 @@
...
+        if (ip == null || ip.length() == 0 || "unknown".equalsIgnoreCase(ip)) {
+            ip = request.getHeader("Proxy-Client-IP");
+        }
+        if (ip == null || ip.length() == 0 || "unknown".equalsIgnoreCase(ip)) {
+            ip = request.getHeader("X-Forwarded-For");
```
bdemers: X-Forwarded-For could be a list:
https://en.wikipedia.org/wiki/X-Forwarded-For#Format

web/src/main/java/org/apache/shiro/web/util/WebUtils.java
```diff
@@ -92,6 +92,32 @@
...
+        }
+        if (ip == null || ip.length() == 0 || "unknown".equalsIgnoreCase(ip)) {
+            ip = request.getHeader("WL-Proxy-Client-IP");
+        }
+        if (ip == null || ip.length() == 0 || "unknown".equalsIgnoreCase(ip)) {
+            ip = request.getHeader("X-Real-IP");
+        }
+
+        if (ip == null || ip.length() == 0 || "unknown".equalsIgnoreCase(ip)) {
```
bdemers: why is `"unknown"` used?

fpapon:
@zhengyanbin Can you fix the comments please? We can add this PR into the next release.

PR #140, opened 2019-05-31
Reduce shiro test logging level to INFO

bdemers:


bdemers:
@fpapon, related to your build cleanup PRs
I have mixed feelings about this, as generally I think extra logging is useful for CI, but in general I would typically just run the test locally and configure the logging i'd need at that point.

Thoughts?

asfgit:
https://builds.apache.org/job/Shiro-pr/99/

fpapon:
@bdemers full agree, we only have to put in `TRACE` or `DEBUG` when we have some errors and help to fix.

fpapon:
@bdemers I think we should also put the root logger to `INFO` in the `samples/guice/src/main/resources/log4j.properties`

asfgit:
https://builds.apache.org/job/Shiro-pr/105/

PR #168, opened 2019-09-22
[SHIRO-684] INI parser keeps escape characters in keys and values

fpapon:
When fixing this issue, I found a bug in the `isCharEscaped` method and the `index` usage.


bdemers:
This seems fine to me, but we might want to add a few more test cases around the escape?

fpapon:
Ok, I will add more tests.

fpapon:
@bdemers I add some tests and also add the escape char check on the value.
The Travis build failed because of a missing jar from the google central maven repository.

PR #177, opened 2019-11-12
Get build working with JDK11

coheigea:
I couldn't get two of the samples working with JDK11, so this patch disables running those two samples with JDK11. I also updated the dependencies in the JAX-RS demo.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -248,6 +248,16 @@
             <plugins>
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-javadoc-plugin</artifactId>
+                    <version>3.1.0</version>
+                    <configuration>
+                        <source>${jdk.version}</source>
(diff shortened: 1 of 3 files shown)


fpapon:
@coheigea Thanks!
There is a JIRA about building on JDK11:
https://issues.apache.org/jira/browse/SHIRO-690

And I add some specific builds on Jenkins to validate:
https://builds.apache.org/view/S-Z/view/Shiro/job/Shiro-pr/173/console

coheigea:
@fpapon Looks like you merged just after I pushed an update to the branch :-) The update was to fix the javadoc stuff for JDK11, so I think we can resolve SHIRO-690.

fpapon:
@coheigea Yes I didn't see your commit before merge, it was just in time :-) 
I will resolve the issue in Jira.
Thanks for fixing this!

PR #185, opened 2019-12-05
[SHIRO-671] Add support for JSR-250 annotations DenyAll, PermitAll, RolesAllowed

bdemers:
Fixes: SHIRO-671

NOTE: this is the first pass, and uncovered a few issues with the differences between the JSR-250 and Shiro Annotations.

- `RolesAllowed` only supports a logical OR, `RequiresRoles` defaults to a logical AND
- `PermitAll` allows any subject (guest or not)
- Per spec there is interaction between the JSR-250 annotations that the Shiro Annotations do NOT have.
  - Each Shiro annotation is processed without knowledge of other annotations, the JSR-250 versions are processed together, for example

```java
    @PermitAll
    class ExamplePermitAtClass {

        @DenyAll
        void expectDenied() {}

        void expectAllowed() {}

        @RolesAllowed({"blah2", "foo"})
        void withRoles() {}
    }
```

To implement this I needed to search out the annotations in `Jsr250MethodInterceptor` without the use of the `AnnotationResolver` component, the issue here is we have a separate implementation for Spring.

We could extend the `AnnotationResolver` interface and add methods that return annotations for a Class, or a method. (the current method does both).

TODO:

- [ ] Make sure Javadoc has been added
- [ ] Add JAX-RS support
- [ ] Add Spring Support
- [ ] Add Guice support

### DIFF
--- a/core/src/main/java/org/apache/shiro/authz/aop/AnnotationsAuthorizingMethodInterceptor.java
+++ b/core/src/main/java/org/apache/shiro/authz/aop/AnnotationsAuthorizingMethodInterceptor.java
@@ -23,6 +23,7 @@ import java.util.Collection;
 
 import org.apache.shiro.aop.MethodInvocation;
 import org.apache.shiro.authz.AuthorizationException;
+import org.apache.shiro.util.ClassUtils;
 
 /**
  * An <tt>AnnotationsAuthorizingMethodInterceptor</tt> is a MethodInterceptor that asserts a given method is authorized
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/apache/shiro/authz/aop/AnnotationsAuthorizingMethodInterceptor.java
```diff
@@ -52,12 +53,16 @@
      * support role and permission annotations.
      */
     public AnnotationsAuthorizingMethodInterceptor() {
-        methodInterceptors = new ArrayList<AuthorizingAnnotationMethodInterceptor>(5);
+        methodInterceptors = new ArrayList<>(8);
```
bdemers: 6 -> 8

fpapon:
retest this please

lprimak (approved):

lprimak:
Superseded by #714

PR #186, opened 2019-12-06
[SHIRO-668] Catch unexpected errors which can lead to oom

fpapon:
This supercedes https://github.com/apache/shiro/pull/71
@bdemers I add a test for the exception catching.
Feedbacks are welcome :)

### DIFF
--- a/core/src/main/java/org/apache/shiro/session/mgt/ExecutorServiceSessionValidationScheduler.java
+++ b/core/src/main/java/org/apache/shiro/session/mgt/ExecutorServiceSessionValidationScheduler.java
@@ -111,8 +111,17 @@ public class ExecutorServiceSessionValidationScheduler implements SessionValidat
         if (log.isDebugEnabled()) {
             log.debug("Executing session validation...");
         }
+        Thread.currentThread().setUncaughtExceptionHandler((t, e) -> {
+            log.error("Error while validating the session, the thread will be stopped and session validation disabled", e);
+            this.disableSessionValidation();
+        });
         long startTime = System.currentTimeMillis();
-        this.sessionManager.validateSessions();
+        try {
+            this.sessionManager.validateSessions();
+        } catch (RuntimeException e) {
+            log.error("Error while validating the session", e);
+            //we don't stop the thread
+        }
         long stopTime = System.currentTimeMillis();
         if (log.isDebugEnabled()) {
             log.debug("Session validation completed successfully in " + (stopTime - startTime) + " milliseconds.");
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/apache/shiro/session/mgt/ExecutorServiceSessionValidationSchedulerTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+    @After
+    public void tearDown() throws Exception {
+        executorServiceSessionValidationScheduler.disableSessionValidation();
+    }
+
+    private class FakeDefaultSessionManager extends DefaultSessionManager {
+        @Override
+        public void validateSessions() {
+            Byte[] fake = null;
```
bdemers: Maybe just throw an exception here to make this more clear?

core/src/main/java/org/apache/shiro/session/mgt/ExecutorServiceSessionValidationScheduler.java
```diff
@@ -111,8 +111,17 @@ public void run() {
         if (log.isDebugEnabled()) {
             log.debug("Executing session validation...");
         }
+        Thread.currentThread().setUncaughtExceptionHandler((t, e) -> {
+            log.error("Error while validating the session, the thread will be stopped and session validation disabled", e);
+            this.disableSessionValidation();
```
bdemers: This seems like a good idea, but I'm not 100% sure on this. 
Are there any uncaught Throwables that could be recovered from?  The JVM doesn't stop, so the application _might_ continue to function just fine

That said, I'm assuming in these cases there is a very low percentage chance of being able to recover, so failing seems like a good option.  Part of me wants to say we should allow the setting of the ExectorService's `UncaughtExceptionHandler` (but default to your impl).  But the other part of me thinks my suggestion is likely over-engineering a situation you cannot recover from anyway.


**NOTE:** take this with a grain of salt, as this is just a quick review

core/src/test/java/org/apache/shiro/session/mgt/ExecutorServiceSessionValidationSchedulerTest.java
fpapon: @bdemers Done ;)

core/src/main/java/org/apache/shiro/session/mgt/ExecutorServiceSessionValidationScheduler.java
fpapon: @bdemers Only exceptions higher than RuntimeException will be catch by the handler. My toughts was just to log a message about the error and trying to disable the executor.

fpapon:
retest this please

PR #200, opened 2020-02-03
Update apsectj-maven-plugin version

bdemers:
Move plugin version and basic config to plugin management


bdemers:
@fpapon trying one more thing for the aspectj build problem

I've only seen the current problem when `release:perform` is run. rerunning the same goals/profile does NOT reproduce the issue.  I figure I'll give the lastest version a try before moving forward.

PR #210, opened 2020-04-01
[SHIRO-530] INI parser does not properly handled backslashes at end o…

bmhm:
…f values

 - Do not skip escape characters for the value (new behaviour demanded by SHIRO-530).
 - rearrange and comment tests to reflect old and new, desired behaviour.

Signed-off-by: Benjamin Marwell <bmarwell@gmail.com>

Alternative to https://github.com/apache/shiro/pull/209

### DIFF
--- a/config/core/src/test/groovy/org/apache/shiro/config/IniTest.groovy
+++ b/config/core/src/test/groovy/org/apache/shiro/config/IniTest.groovy
@@ -73,6 +73,22 @@ public class IniTest {
     }
 
     @Test
+    public void testBackslash() {
+        String test = "Truth=Beauty\\\\";
+        Ini ini = new Ini();
+        ini.load(test);
+
+        assertNotNull(ini.getSections());
+        assertEquals(1, ini.getSections().size());
+
+        Ini.Section section = ini.getSections().iterator().next();
+        assertEquals(Ini.DEFAULT_SECTION_NAME, section.getName());
+        assertFalse(section.isEmpty());
+        assertEquals(1, section.size());
+        assertEquals("Beauty\\\\", section.get("Truth"));
+    }
+
+    @Test
     public void testSplitKeyValue() {
         String test = "Truth Beauty";
         String[] kv = Ini.Section.splitKeyValue(test);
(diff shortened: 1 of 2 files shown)


fpapon:
@bmhm sounds good!

fpapon:
@bdemers can you review please just before I merge it?

fpapon:
retest this please

config/core/src/test/groovy/org/apache/shiro/config/IniTest.groovy
```diff
@@ -119,24 +135,40 @@ public class IniTest {
         assertEquals("Truth", kv[0]);
         assertEquals("Beauty", kv[1]);
 
+        // Escape characters are to be removed from the key.
+        // This is different behaviour compared to the XML config.
```
bdemers, line 139: This seems odd, but I guess it's existing functionality

config/core/src/test/groovy/org/apache/shiro/config/IniTest.groovy
```diff
@@ -119,24 +135,40 @@ public class IniTest {
...
+        // SHIRO-530: Keep backslashes in value.
+        test = "Truth=Beau\\ty";
         kv = Ini.Section.splitKeyValue(test);
         assertEquals("Truth", kv[0]);
-        assertEquals("Beauty", kv[1]);
+        assertEquals("Beau\\ty", kv[1]);
 
-        test = "Truth=Beau\\ty";
+        // SHIRO-530: Keep backslashes in value.
+        test = "Truth=Beauty\\";
```
bdemers, line 152: shouldn't this one wrap? and the next line would be included?
i.e. something like:
```ini
Truth=Beauty\
more text here
```
Which would be something like `Truth=Beatuty\\\nmore text here`

fpapon:
I think the complexity is that we can have some regex or URI in the value, so we have to deal correctly with the backslash

config/core/src/test/groovy/org/apache/shiro/config/IniTest.groovy
bmhm: Yes, I kept that. They just disappear magically.

config/core/src/test/groovy/org/apache/shiro/config/IniTest.groovy
bmhm: No, it should not, because we are calling the `splitKeyValue` method. The backslashes at the end are resolved in the caller method. When arriving at this point, the wrapping would have already occured. Took me a while, too, to figure this one out. But I did not bother to add more comments. ;-)

bmhm: @bdemers see here: https://github.com/apache/shiro/blob/b50b829a285b106666b688a4b69a3ebee94f51b4/config/core/src/main/java/org/apache/shiro/config/Ini.java#L609-L631

Please check if I did not mis-read it.

fpapon:
@bmhm Thanks for the great work ;)

PR #224, opened 2020-05-06
[SHIRO-770] Remove base64 implementation, keep UTF-8 codec by default.

bmhm:
For v2.0.0, we probably do not want to maintain our own Base64 implementation.

Still, the shiro implementation defaulted to UTF-8 for decoding strings, which seems to be a sane default.

However, the `CryptoException` became unreachable from the Base64 class. As this is not cryptography anyway, I decided to replace it with the `IllegalArgumentException` thrown by `java.util.Base64`'s methods.


fpapon:
Build is broken because of jacip maven plugin for api incompatibility but it's normal as we move on 2.0.0 and we want to broke this.
I will deactivate this.

bdemers:
> Build is broken because of jacip maven plugin for api incompatibility but it's normal as we move on 2.0.0 and we want to broke this.
> I will deactivate this.

+1, the config probably needs to change as a major version change in the pom _should_ allow for breaking API changes.
But, for exactly that same reason, we don't need to be running japicmp right now anyway.

Once we get closer to a beta for date for 2.x, we can generate a japicmp report and list out the API changes.

bdemers:
@bmhm great change! PRs that remove code are my favorite 😄

fpapon:
retest this please

fpapon:
@bmhm can you create a Jira for this issue?

bmhm:
> @bmhm can you create a Jira for this issue?

SHIRO-770 - https://issues.apache.org/jira/browse/SHIRO-770

PR #232, opened 2020-05-09
[SHIRO-768] Remove the shiro-all module

fpapon:


fpapon:
@bdemers can you review please?

PR #236, opened 2020-05-12
[SHIRO-679] Shiro modules have split packages

fpapon:
**Modules**

* config/ogdl:
  - create package org.apache.shiro.config.ogdl and move everything into

* core: 
  - rename package config to ini (only deprecated classes into) to avoid conflict with config-core module package

* crypto:
  - cipher: create package org.apache.shiro.crypto.cipher and move everything into

* lang: 
  - create package org.apache.shiro.lang and move everyting into

* support/feature:
  - redefine shiro-core feature with all needed bundles

**Global**

  - export-package only provided by the bundle, avoid using org.apache.shiro*

Tested on Apache Karaf 4.2.8


steinarb:
LGTM

fpapon:
Thanks guys for the review, I merge it and we can move forward on 2.0.0!

PR #239, opened 2020-06-14
[SHIRO-780] NOTICE files of shiro components don't match NOTICE in so…

fpapon:
…urce code repository

The apache-jar-resource-bundle generate default NOTICE file from velocity template, so we have to override this default NOTICE file by the root project one.

### DIFF
--- a/cache/src/main/resources/META-INF/NOTICE
+++ b/cache/src/main/resources/META-INF/NOTICE
@@ -1,0 +1,15 @@
+Apache Shiro
+Copyright 2008-2020 The Apache Software Foundation
+
+This product includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
+
+The implementation for org.apache.shiro.util.SoftHashMap is based 
+on initial ideas from Dr. Heinz Kabutz's publicly posted version 
+available at http://www.javaspecialists.eu/archive/Issue015.html,
+with continued modifications.  
+
+Certain parts (StringUtils, IpAddressMatcher, etc.) of the source
+code for this  product was copied for simplicity and to reduce
+dependencies  from the source code developed by the Spring Framework
+Project  (http://www.springframework.org).
--- a/config/core/src/main/resources/META-INF/NOTICE
+++ b/config/core/src/main/resources/META-INF/NOTICE
@@ -1,0 +1,15 @@
+Apache Shiro
+Copyright 2008-2020 The Apache Software Foundation
+
+This product includes software developed at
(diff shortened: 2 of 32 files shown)


bdemers:
We should probably filter these files in if/when needed, possibly just in the core/lang module (and if they get repackaged in the `all` module). Or copy a single static NOTICE file in from the root (instead of duplicating it)

Ideas:

A side note, we can also use the `org.jasig.maven:maven-notice-plugin` plugin to generate a notice
https://github.com/paseto-toolkit/jpaseto/blob/master/pom.xml#L239-L251
(**NOTE:** the usage above differs from the ASF requirements)


Another trick is to define a `root.dir` property: in the parent pom:
```xml
<root.dir>${session.executionRootDirectory}</root.dir>
```
And include the resource as `${root.dir}/NOTICE`
But... depending on how maven multi-modules are being used in the project it _might_ be easier to reference them as `../../NOTICE` instead.

bdemers:
NOTE:  I'm not against maintaining the NOTICE by hand. It might be easier to handwrite notice files in a couple of the modules were needed, and merge them into the project's root NOTICE file.

bdemers:
@fpapon I'm not sure if any of these suggestions work around the template issue either)

fpapon:
@bdemers the problem is the apache pom parent definition with the apache-jar-resource-bundle, it generate the NOTICE file from Apache template.
I'm agree that it's not nice to maintain... I will made some tests, may be by copying the root NOTICE file in target folder during process cycle. I'll keep you posted ;)

bdemers:
OHHH, I see now.

I didn't realize where that was coming from.  We can probably get away with the template for most modules.  And only include a custom NOTICE (just copied in as you have) in the modules were needed (for example `lang`)

But I'm not sure if/how the bundle plugin plays into this, because IIRC the bundle plugin repacks some of the dependencies?

We could also "skip" this plugin's execution and make sure the LICENSE & NOTICE files are included by other means.

Anyway, I'm not trying to create more work just to work around a NOTICE bundling issue.  If you can simplify it, great, if not, what you have in this PR is legally correct (which is more important)

fpapon:
Ok, no problem ;) 
I will try to config the apache plugin.

fpapon:
@bdemers I'm not sure to have more time to work on this for the 1.5.x/1.6 release
I propose to merge this one as-is and work on your proposal for the plugin adjust on the 2.0.0 release
Thoughts?

bdemers:
Works for me

PR #240, opened 2020-06-14
[SHIRO-778] onInit method on AuthenticatingRealm is called twice

fpapon:

### DIFF
--- a/config/ogdl/src/main/java/org/apache/shiro/config/ogdl/ReflectionBuilder.java
+++ b/config/ogdl/src/main/java/org/apache/shiro/config/ogdl/ReflectionBuilder.java
@@ -287,10 +287,15 @@ public class ReflectionBuilder {
             }
 
             processor.execute();
-        }
 
-        //SHIRO-413: init method must be called for constructed objects that are Initializable
-        LifecycleUtils.init(objects.values());
+            //SHIRO-778: onInit method on AuthenticatingRealm is called twice
+            objects.keySet().stream()
+                    .filter(key -> !kvPairs.containsKey(key))
+                    .forEach(key -> LifecycleUtils.init(objects.get(key)));
+        } else {
+            //SHIRO-413: init method must be called for constructed objects that are Initializable
+            LifecycleUtils.init(objects.values());
+        }
 
         return objects;
     }
--- a/config/ogdl/src/test/groovy/org/apache/shiro/config/ogdl/ReflectionBuilderTest.groovy
+++ b/config/ogdl/src/test/groovy/org/apache/shiro/config/ogdl/ReflectionBuilderTest.groovy
@@ -18,6 +18,7 @@
  */
 package org.apache.shiro.config.ogdl
 
+import org.apache.shiro.config.ogdl.beans.InitCountBean
 import org.apache.shiro.lang.codec.Base64
 import org.apache.shiro.lang.codec.CodecSupport
 import org.apache.shiro.lang.codec.Hex
@@ -25,6 +26,8 @@ import org.apache.shiro.config.ConfigurationException
 import org.apache.shiro.config.Ini
 import org.apache.shiro.config.ogdl.event.BeanEvent
 import org.junit.Test
+
+import java.util.concurrent.ConcurrentHashMap
 
 import static org.junit.Assert.*
 import static org.hamcrest.Matchers.*
@@ -638,6 +641,37 @@ class ReflectionBuilderTest {
         assertThat beanOne.stringList, allOf(hasItem(System.getProperty("os.name")), hasItem(System.getProperty("os.arch")), hasSize(2))
 
         assertNotNull(beanMap.get("two"))
+    }
+
+    @Test
+    void testNotMultipleInitialization() {
+        // given
+        Map<String, String> defs = new ConcurrentHashMap<>()
+        defs.put("initcountbean", InitCountBean.getCanonicalName())
+        ReflectionBuilder builder = new ReflectionBuilder()
+
+        // when
+        builder.buildObjects(defs)
+
+        // then
+        assertEquals(1, InitCountBean.getInitCount())
+        InitCountBean.resetCount()
+    }
+
+    @Test
+    void testNotMultipleInitializationWithNullFirst() {
+        // given
+        Map<String, String> defs = new ConcurrentHashMap<>()
+        defs.put("initcountbean", InitCountBean.getCanonicalName())
+        ReflectionBuilder builder = new ReflectionBuilder()
+
+        // when
+        builder.buildObjects(null)
+        builder.buildObjects(defs)
+
+        // then
+        assertEquals(1, InitCountBean.getInitCount())
+        InitCountBean.resetCount()
     }
 
     void assertInstantiatedEvents(String name, Map<String, ?> objects, int expected) {
--- a/config/ogdl/src/test/java/org/apache/shiro/config/ogdl/beans/InitCountBean.java
+++ b/config/ogdl/src/test/java/org/apache/shiro/config/ogdl/beans/InitCountBean.java
@@ -1,0 +1,53 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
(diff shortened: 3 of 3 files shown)


config/ogdl/src/main/java/org/apache/shiro/config/ogdl/ReflectionBuilder.java
```diff
@@ -53,6 +43,19 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
```
bmhm, line 58: is the reordering necessary for such a small change? Or do we want to change the ordering in general for v2.0.0?

config/ogdl/src/main/java/org/apache/shiro/config/ogdl/ReflectionBuilder.java
```diff
@@ -290,7 +293,9 @@ private String parseBeanId(String lhs) {
         }
 
         //SHIRO-413: init method must be called for constructed objects that are Initializable
-        LifecycleUtils.init(objects.values());
+        //SHIRO-778: onInit method on AuthenticatingRealm is called twice
+        objects.keySet().stream().filter(key -> !Optional.ofNullable(kvPairs).orElseGet(HashMap::new).keySet().contains(key))
+                .forEach(key -> LifecycleUtils.init(objects.get(key)));
```
bmhm, line 298: Having worked with a lot of lambdas, two suggestions here:
* Line break after `.stream()` and  each operation (after `.filter()` and `.forEach()` in this case).
* refactor out the `.filter()` contents. It is much easier to read a method name here than to grasp the contents of the whole filter operation and where parentheses actually ends.

config/ogdl/src/main/java/org/apache/shiro/config/ogdl/ReflectionBuilder.java
fpapon: I didn't watch this, it's my IDEA settings that reordered the import, I will revert it.

config/ogdl/src/main/java/org/apache/shiro/config/ogdl/ReflectionBuilder.java
fpapon: ok for the break line. However for the refactor out the filter() content, I'm not fully agree, the content is short.

bmhm: Ok, I got an better idea anyway.
At the moment, you are creating a new map for each key of kvPairs is empty! That's a lot of overhead!

At least move this part to its own statement on the previous line:
`Optional.ofNullable(kvPairs).orElseGet(HashMap::new)`.

This will avoid creating 10 new hashmaps of you have 10 keys in objects. And it's much easier to read the lambda (if !map.contains)

config/ogdl/src/main/java/org/apache/shiro/config/ogdl/ReflectionBuilder.java
```diff
@@ -290,7 +293,9 @@ private String parseBeanId(String lhs) {
         }
 
         //SHIRO-413: init method must be called for constructed objects that are Initializable
-        LifecycleUtils.init(objects.values());
+        //SHIRO-778: onInit method on AuthenticatingRealm is called twice
+        objects.keySet().stream().filter(key -> !Optional.ofNullable(kvPairs).orElseGet(HashMap::new).keySet().contains(key))
```
bmhm, line 297: I see another issue with this line. You are using `HashMap::new`. Let's just pretend you are following my comment to remove the overhead of instantiating the map for each key. We might have a line like this:

`Map kvPairsToCheck = Optional.ofNullable(kvPairs).orElseGet(HashMap::new)`.

If you expect the devs to modify this map in a later commit, please do not use `HashMap::new` which is prone to `ConcurrentModificationException`. A lot of apache projects prohibit it's use via checkstyle for this reason.

If you do not expect devs to modify this map in later commits, you might as well use [`Collections::emptyMap`](https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#emptyMap--), which will hint the dev that this map is going to stay empty (because `Collections` will usually return unmodifiable maps). This way, if someone tries to add content to this map to a later release, he/she will receive an `UnsupportedOperationException` which will make them think about why this was unmodifiable in the first place.
As this only kicks in if the original map `kvPairs` is empty, this seems a sane choice.

Then, `Collections::emptyMap` might receive performance updates in future releases, while `HashMap::new` is less likely to.

bmhm: Oh, btw, the javadoc states: 
> Implementations of this method need not create a separate Map object for each call. Using this method is likely to have comparable cost to using the like-named field. (Unlike this method, the field does not provide type safety.)

bmhm: Oh, and of course, instead of calling 
`Optional.ofNullable(kvPairs).orElseGet(HashMap::new).keySet()`
when refactoring into it's own line you could also use this:
`Set knownKeys = Optional.ofNullable(kvPairs).map(Map::keySet).orElseGet(Collections::emptySet)`.

I mean, the filter is about the set containing a key, right? Or if you avoid mapping into a new collection, you can use containsKey instead: 
`.filter(key -> !map.containsKey(key))`.
That's probably even better because it avoids creation of a new collection and is even shorter to read and easier to understand on first sight.

fpapon: Let me make some polish ;)

fpapon:
@bdemers I made some polish, I think it's better now.

config/ogdl/src/main/java/org/apache/shiro/config/ogdl/ReflectionBuilder.java
```diff
@@ -287,10 +287,15 @@ private String parseBeanId(String lhs) {
             }
 
             processor.execute();
-        }
 
-        //SHIRO-413: init method must be called for constructed objects that are Initializable
-        LifecycleUtils.init(objects.values());
+            //SHIRO-778: onInit method on AuthenticatingRealm is called twice
+            objects.keySet().stream()
+                    .filter(key -> !kvPairs.keySet().contains(key))
```
bmhm, line 293: I think it is not necessary to call `.keySet()` when you can just use `.containsKey(key)`. This will create a new view each time, the latter might also be a little bit easier to read.

fpapon: Right

bdemers:
Looks good,! it would be great to have a test for this one. (possibly just tweaking the existing test `InitializableBean` to count the number of times `init()` was called)

bmhm:
> Looks good,! it would be great to have a test for this one. (possibly just tweaking the existing test `InitializableBean` to count the number of times `init()` was called)

Or just create a mock objects and see if `values()` was called using an appropriate `verify()` method. This way you would not need to change production code for tests.

config/ogdl/src/main/java/org/apache/shiro/config/ogdl/ReflectionBuilder.java
```diff
@@ -287,10 +287,15 @@ private String parseBeanId(String lhs) {
...
-        //SHIRO-413: init method must be called for constructed objects that are Initializable
-        LifecycleUtils.init(objects.values());
+            //SHIRO-778: onInit method on AuthenticatingRealm is called twice
+            objects.keySet().stream()
+                    .filter(key -> !kvPairs.containsKey(key))
+                    .forEach(key -> LifecycleUtils.init(objects.get(key)));
+        } else {
+            //SHIRO-413: init method must be called for constructed objects that are Initializable
+            LifecycleUtils.init(objects.values());
+        }
```
bmhm, line 298: This else branch will cause calling initialization witce.
Consider 
```
        builder.buildObjects(defs);
        builder.buildObjects(null);
```

bmhm: https://github.com/fpapon/shiro/pull/1

fpapon: I'm not sure that users can use the buildObjects, it's not a part of the public api so I can't see the case where this method will be call several time at startup.
But if we need to add a control, may be we can add a boolean method in the Initializable interface to define if the bean has been initialized.
@bdemers @bmhm thoughts?

config/ogdl/src/test/groovy/org/apache/shiro/config/ogdl/ReflectionBuilderTest.groovy
```diff
@@ -640,6 +643,20 @@ class ReflectionBuilderTest {
...
+        defs.put("initcountbean", InitCountBean.getCanonicalName())
+        ReflectionBuilder builder = new ReflectionBuilder()
+
+        // when
+        builder.buildObjects(defs);
+
+        // then
+        assertEquals(1, InitCountBean.getInitCount())
+    }
+
```
bmhm, line 659: Can you add a test where `buildObjects(null)` is called FIRST? Because that actually happens in Shiro.

bmhm:
LGTM :+1: 
@bdemers WDYT?

PR #242, opened 2020-06-19
[SHIRO-610] Allways create resolver for non-empty IniWebEnvironment

tbrugz:

### DIFF
--- a/web/src/main/java/org/apache/shiro/web/env/IniWebEnvironment.java
+++ b/web/src/main/java/org/apache/shiro/web/env/IniWebEnvironment.java
@@ -263,20 +263,13 @@ public class IniWebEnvironment extends ResourceBasedWebEnvironment implements In
         Ini ini = getIni();
 
         if (!CollectionUtils.isEmpty(ini)) {
-            //only create a resolver if the 'filters' or 'urls' sections are defined:
-            Ini.Section urls = ini.getSection(IniFilterChainResolverFactory.URLS);
-            Ini.Section filters = ini.getSection(IniFilterChainResolverFactory.FILTERS);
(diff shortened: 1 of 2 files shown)


web/src/main/java/org/apache/shiro/web/env/IniWebEnvironment.java
```diff
@@ -263,20 +263,13 @@ protected FilterChainResolver createFilterChainResolver() {
         Ini ini = getIni();
 
         if (!CollectionUtils.isEmpty(ini)) {
-            //only create a resolver if the 'filters' or 'urls' sections are defined:
```
bmhm, line 266: I'd like to see a comment here, if you don't mind:
`// SHIRO-610: Allow creation of FilterChainResolver even without urls and filters sections.`

fpapon: > I'd like to see a comment here, if you don't mind:
> `// SHIRO-610: Allow creation of FilterChainResolver even without urls and filters sections.`

@bmhm I'm not a big fan to add the jira related informations, we already have the git commit history for this.

bmhm: Alright. I was requesting this because we had security-related code changes which could have been prevented by such comments. Why? Because  you do not always look into the history before accepting code or changing code.

fpapon: @bmhm ok, so may be adding the related CVE id?

tbrugz:
@fpapon @bmhm Let me know if I should add something else to the PR

fpapon:
@bmhm can you review please?

PR #259, opened 2020-10-15
Update AbstractContainerIT to allow for HTTPS connections      Using a pre-generated keystore (1.6.x)

bdemers:


integration-tests/support/src/main/java/org/apache/shiro/testing/web/AbstractContainerIT.java
```diff
@@ -109,6 +128,26 @@ protected WebAppContext createdWebAppContext() throws Exception {
...
+        sslContextFactory.setKeyStorePath(TEST_KEYSTORE_PATH.getAbsolutePath());
+        sslContextFactory.setKeyStorePassword(TEST_KEYSTORE_PASSWORD);
+        sslContextFactory.setKeyManagerPassword(TEST_KEYSTORE_PASSWORD);
+
+        HttpConfiguration https = new HttpConfiguration();
+        https.addCustomizer(new SecureRequestCustomizer());
+
+        final ServerConnector httpsConnector = new ServerConnector(
+                server,
+                new SslConnectionFactory(sslContextFactory, HttpVersion.HTTP_1_1.asString()),
```
bmhm, line 146: Any reason why we would explicitly don't want http/2 (e.g. for java 9+ builds)?

integration-tests/support/src/main/java/org/apache/shiro/testing/web/AbstractContainerIT.java
```diff
@@ -150,4 +193,30 @@ public static void stopContainer() {
...
+    // Dealing with a keystore is NOT fun, it's easier to script one with the keytool
+    // see src/main/resources/createKeyStore.sh for more info
+    private static File setupKeyStore() {
+        try {
+            Path outKeyStoreFile = File.createTempFile("test-keystore", "jks").toPath();
+            URL keyStoreResource = Thread.currentThread().getContextClassLoader().getResource("test-keystore.jks");
+            Files.copy(keyStoreResource.openStream(), outKeyStoreFile, StandardCopyOption.REPLACE_EXISTING);
+            File keyStoreFile = outKeyStoreFile.toFile();
+
+            // _most_ clients will pick up the ssl keystore this way, so just set SSL properties
```
bmhm, line 214: I think you can remove the “most". It doesn't add valuable information of you don't specify which don't.
All JAX-RS implementations and all Apache httpClients will pick them up.

bmhm (changes requested):

integration-tests/support/src/main/resources/createKeyStore.sh
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
bdemers: This script isn't run at build time, it just exists as a source for the keystore and pem as part of this commit.
I'd rather create a keystore on the fly as needed by tests, but this was easier in the short term.

integration-tests/support/src/main/java/org/apache/shiro/testing/web/AbstractContainerIT.java
bdemers: +💯 
I gave it a quick try, but other jetty modules need to be included to support http/2. That can be addressed in the future

samples/web/src/test/java/org/apache/shiro/test/WebAppContainerIntegrationIT.java
```diff
@@ -31,14 +31,14 @@
 import java.io.IOException;
 import java.net.MalformedURLException;
 
-public class ContainerIntegrationIT extends AbstractContainerIT {
+public class WebAppContainerIntegrationIT extends AbstractContainerIT {
 
     protected final WebClient webClient = new WebClient();
```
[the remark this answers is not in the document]
bdemers: In my first pass through I did that :)
But the keystore is already available to the JVM, so the validation just works.  I do agree though, validating the TLS on the client side for the ITs doesn't add additional value (for what we are testing).

My thought is, leave the default config when possible, and enable proper validation in the test clients when it's easy, and then disable it when it's not (with a big warning about NOT doing this in production, etc)

bmhm (approved):

PR #273, opened 2020-12-31
[SHIRO-290] Implement BCrypt (and possibly other unix crypt formats like scrypt)

bmarwell:
- [X] Create AbstractCryptClass for Unix-Crypt-based algorithms (e.g. bcrypt, script, argon2)
- [X] Move some of the password matching logic into the Hash classes
    Reviwers: Review method placement and HAsh class description.
- [X] removed hashrequest
- [X] API change: made salt not-nullable.
    Additional constructor is supplied for hashing without or with
    default salt, the former and other methods/fields using
    SimpleByteSource.empty().
    Reviewers: Pay attention to method logic, so no empty salt is
    being used where a former `null` value would have created a
    new, random salt.
- [X] Modified tests to not expect exceptions in certain cases.
- [X] Modified tests to not expect passwordService calls when supplying an
    existing hash.
- [x] Fix Javadocs
- [x] Fix Hasher utility
- [x] Add Argon2 algorithms
- [x] Deprecate old non-KDF hash classes
- [ ] ~~Also print warnings when encountering the old hashes~~
- [x] What should `getBytes()` return? There is no overridden Javadoc. Currently, only the hashed data is returned, but maybe it was meant to also return the hash here? If not, why even implement `CodecSupport`? This needs to be clarified before merging.
   => This is more important if we want to retain Hex and Base64 formats. They need all parameters encoded.
 - [x] BCrypt iterations vs cost: What do we return for `getIterations()`? Currently I return the cost. But we could also return the real iterations (which is `2^cost`) it that made sense.
 - [X] ~~How to store argon2 parameters in Shiro1CryptFormat? Maybe create a Shiro2 CryptFormat?
   Argon2 outputs this: `$argon2i$v=19$m=65536,t=2,p=4$c29tZXNhbHQ$RdescudvJCsgt3ub+b+dWRWJTmaaJObG`.
  The Shiro1 crypt format only has: prefix(shiro1), algo, iterationcount, salt, hash.
  We would also need version and parameters or so, which could be empty for others.
  The current working copy uses a comma separated list for the iterationCount field, but I am not happy with that.~~
- [X] Shiro2CryptFormat (see comment above)?
- [X] revert Shiro1CryptFormat modifications?
- [X] ~~Can we drop Hex and Base64 Formats?~~ => Can be done in another PR.
- [X] Add more tests for bcrypt and argon2.
- [ ] ~~Tune Argon2 default parameters. Target computation time on servers: ~ 0.1s as per RFC recommendation.~~
- [ ] ~~Add scrypt. It is better than bcrypt.~~ => Another PR.
- [x] What does `DefaultPasswordService.testStringComparisonWhenNotUsingAParsableHashFormat` actually test? It only works if specific parameters are set.

## API Changes

- Changed default DefaultPasswordService.java algorithm to "Argon2id".
- `PasswordService.encryptPassword(Object plaintext)` will now throw
       a NullPointerException on `null` parameter. It was never specified how this method would behave.
- made salt not-nullable.

---

 - [x] Run `mvn clean install apache-rat:check` to make sure basic checks pass. A more thorough check will be performed on your pull request automatically.


fpapon (changes requested):
@bmarwell we can add @since 2.0 to the new classes and the OpenBSD64 part come from another project so I think we have to mention it in the NOTICE file.

bmarwell:
About `HexFormat` and `Base64Format`:

> `HexFormat.java`
> HashFormat that outputs only The hash's digest bytes in hex format. It does not print out anything else (salt, iterations, etc). This implementation is mostly provided as a convenience for command-line hashing.

and

> `Base64Format.java`
> HashFormat that outputs only the hash's digest bytes in Base64 format. It does not print out anything else (salt, iterations, etc). This implementation is mostly provided as a convenience for command-line hashing.

my +1 for dropping them:
1. for 2.0.0  mark them as deprecated and log warnings when being used.
2. for 2.1.0 throw exceptions when being used
3. for 2.2.0 complete removal

core/pom.xml
```diff
@@ -101,6 +101,14 @@
             <groupId>org.apache.shiro</groupId>
             <artifactId>shiro-crypto-hash</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.shiro</groupId>
+            <artifactId>shiro-crypto-support-hashes-argon2</artifactId>
```
bdemers, line 106: nit: he groupId could be `org.apache.shiro.hashes` and artifactId: `shiro-crypto-argon2`
Or, `org.apache.shiro.crypto:shiro-hashes-argon2`

crypto/hash/src/main/java/org/apache/shiro/crypto/hash/ConfigurableHashService.java
```diff
@@ -18,44 +18,20 @@
...
-
     /**
      * Sets the name of the {@link java.security.MessageDigest MessageDigest} algorithm that will be used to compute
      * hashes.
      *
      * @param name the name of the {@link java.security.MessageDigest MessageDigest} algorithm that will be used to
      *             compute hashes.
      */
-    void setHashAlgorithmName(String name);
+    void setDefaultAlgorithmName(String name);
```
bdemers, line 35: nit: might need to tweak the javadoc to go with this, I'm assuming this is the alg used if the hash cannot be determined?

crypto/hash/src/main/java/org/apache/shiro/crypto/hash/ConfigurableHashService.java
```diff
@@ -18,44 +18,20 @@
...
 
-    /**
-     * Sets a source of randomness used to generate public salts that will in turn be used during hash computation.
-     *
-     * @param rng a source of randomness used to generate public salts that will in turn be used during hash computation.
-     */
-    void setRandomNumberGenerator(RandomNumberGenerator rng);
```
bdemers, line 60: +1 for removing this from the public API.

crypto/hash/src/main/java/org/apache/shiro/crypto/hash/DefaultHashService.java
```diff
@@ -123,222 +71,52 @@ public DefaultHashService() {
...
-     * to protect the integrity of the private salt (without a public salt, the private salt would be exposed as-is,
-     * which is undesirable).
-     *
-     * @param generatePublicSalt whether or not a public salt should be randomly generated and used to compute a hash
-     *                           if a {@link HashRequest} does not specify a salt.
-     */
-    public void setGeneratePublicSalt(boolean generatePublicSalt) {
-        this.generatePublicSalt = generatePublicSalt;
+    public void setRandom(Random random) {
+        this.random = random;
```
bdemers, line 120: If possible make the random getter/setter package private (I'm guessing it's only used for testing?)

crypto/hash/src/main/java/org/apache/shiro/crypto/hash/HashProvider.java
```diff
@@ -0,0 +1,60 @@
...
+        ServiceLoader<HashSpi<? extends Hash>> hashSpis = load();
+
+        return StreamSupport.stream(hashSpis.spliterator(), false)
+                .filter(hashSpi -> hashSpi.getImplementedAlgorithms().contains(algorithmName))
+                .findAny();
+    }
+
+    @SuppressWarnings("unchecked")
+    private static ServiceLoader<HashSpi<? extends Hash>> load() {
+        return (ServiceLoader<HashSpi<? extends Hash>>) (Object) ServiceLoader.load(HashSpi.class);
```
bdemers, line 57: In JJWT, we ran into some issues with class loaders on various environments, IIRC depending on the env: vanilla java, JPMS, spring, OSGi, the actual classloader used could be different, so we ended up doing something like this:
https://github.com/jwtk/jjwt/blob/master/impl/src/main/java/io/jsonwebtoken/impl/lang/Services.java

(basically checking the three common classloaders)
I don't think this is blocking, as it shouldn't change the API of this class, but it's something to think about

crypto/hash/src/main/java/org/apache/shiro/crypto/hash/Md2Hash.java
```diff
@@ -30,10 +30,13 @@
  * techniques and how the overloaded constructors function.
  *
  * @since 0.9
+ * @deprecated and will throw exceptions since 2.0.0, to be removed in 2.1.0.
```
bdemers, line 33: Rip anything out in 2.0, removing anything is technically a breaking change. (it's easy to add back in later if/when needed)

bdemers: we can also mark them deprecated now, and then remove all the deprecated bits before the official 2.0 release if you are on the fence about anything.

Another idea, is we could nag users: print a log message: "WARN: Seriously, stop using MD2, here is why <insert some URL here>" 👺

crypto/support/hashes/bcrypt/src/main/java/org/apache/shiro/crypto/support/hashes/bcrypt/OpenBSDBase64.java
```diff
@@ -0,0 +1,194 @@
...
+     *  distributed under the License is distributed on an "AS IS" BASIS,
+     *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+     *  See the License for the specific language governing permissions and
+     *  limitations under the License.
+     */
+
+    /**
+     * A mod of Square's Okio Base64 encoder
+     * <p>
+     * Original author: Alexander Y. Kleymenov
```
bdemers, line 74: If not already make sure this gets added to the NOTICE for this module

crypto/support/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+        <version>2.0.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>support</artifactId>
+    <name>Apache Shiro :: Cryptography :: Support</name>
+    <packaging>pom</packaging>
+
+    <modules>
+        <module>hashes</module>
```
bdemers, line 38: nit: you can remove the extra empty parent poms by doing something like:
```xml
<modules>
  <module>hashes/bcrypt</module>
  <module>hashes/argon2</module>
</modules>
```

tools/hasher/pom.xml
```diff
@@ -50,7 +50,11 @@
         </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-simple</artifactId>
+            <artifactId>slf4j-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
```
bdemers, line 57: nice cleanup 😄

crypto/hash/src/main/java/org/apache/shiro/crypto/hash/ConfigurableHashService.java
bmarwell: Yes, the javadocs are the unchecked checkbox ;) Next commit is on the way.

The default is being used when a `HashRequest` does not carry an algorithm name. Its old behaviour/semantics was at least similar, but I was not looking into making it a 100% compatible here.

crypto/hash/src/main/java/org/apache/shiro/crypto/hash/DefaultHashService.java
bmarwell: Not even for testing. Removing it until we need it.

crypto/hash/src/main/java/org/apache/shiro/crypto/hash/Md2Hash.java
bmarwell: +1 for ripping out. We should nag the users in 1.7+ then.
And not just for MD2. For all iteration+salt+pepper based hashes, even if there is no alternative in shiro 1. Seriously.

tools/hasher/pom.xml
bmarwell: It is not so funny actually. See the chat in slack. To be able test things written to `System.out`, we need `logback-classic`, as the `simple-logger` cannot be intercepted.
Romain suggested to refactor the hasher utility a bit further, and I would like to do that in another PR.

crypto/hash/src/main/java/org/apache/shiro/crypto/hash/HashProvider.java
bmarwell: What bugs me most is the ugly casting to object and back. But with what I saw from your `jjwt` example, adding a parameter might help with that!

bmarwell: I removed the generic type.

PR #286, opened 2021-03-13
[SHIRO-812] Key value separator in config is broken with escape char

fpapon:

### DIFF
--- a/config/core/src/main/java/org/apache/shiro/config/Ini.java
+++ b/config/core/src/main/java/org/apache/shiro/config/Ini.java
@@ -579,7 +579,7 @@ public class Ini implements Map<String, Ini.Section> {
                 char c = line.charAt(i);
 
                 if (buildingKey) {
-                    if (isKeyValueSeparatorChar(c) && !isCharEscaped(line, i)) {
+                    if (isKeyValueSeparatorChar(c) && !isCharEscaped(line, i) && !isCharEscaped(line, i-1)) {
                         buildingKey = false;//now start building the value
                     } else if (!isCharEscaped(line, i)){
                         keyBuffer.append(c);
--- a/config/core/src/test/groovy/org/apache/shiro/config/IniTest.groovy
+++ b/config/core/src/test/groovy/org/apache/shiro/config/IniTest.groovy
@@ -159,6 +159,11 @@ public class IniTest {
         kv = Ini.Section.splitKeyValue(test);
         assertEquals("Truth", kv[0]);
         assertEquals("\\ Beauty\\", kv[1]);
+
+        test = "cn\\=TheSpecial_GroupName,ou\\=groups,dc\\=example,dc\\=com = *:*"
+        kv = Ini.Section.splitKeyValue(test)
+        assertEquals("cn=TheSpecial_GroupName,ou=groups,dc=example,dc=com", kv[0])
+        assertEquals("*:*", kv[1])
     }
 
     /**


fpapon:
Fix this https://github.com/apache/shiro/commit/0c424d37e5d542844c9ed65cf962498211d2334b

fpapon:
I have one test failed but do we want this to be possible?
```
    @Test
    public void testSplitKeyValueEscapedEquals()  {
        String test = "Truth\\=Beauty";
        String[] kv = Ini.Section.splitKeyValue(test);
        assertEquals("Truth", kv[0]);
    }
```

bmarwell:
> I have one test failed but do we want this to be possible?
> 
> ```
>     @Test
>     public void testSplitKeyValueEscapedEquals()  {
>         String test = "Truth\\=Beauty";
>         String[] kv = Ini.Section.splitKeyValue(test);
>         assertEquals("Truth", kv[0]);
>         assertEquals("Beauty", kv[1]);
>     }
> ```

For 2.0: no. For me the equals is escaped.

bmarwell (approved):

charliemblack:
I tested out the change with an LDAP integration I have with Apache Geode.   The fix works like a champ!   Thanks for doing this.

If I could request this to be back-ported to 1.7.x since that is what Apache Geode is on.

bdemers (approved):

bmarwell:
> If I could request this to be back-ported to 1.7.x since that is what Apache Geode is on.

I am not sure we can backport this, as this is a breaking change. See https://github.com/apache/shiro/pull/286#issuecomment-798887811.
OTOH, I cannot imagine someone has ever used a kv-pair like this intentionally.

@fpapon WDYT?

bdemers:
IIRC, @bmarwell added some tests with the current behavior in: b50b829a285b106666b688a4b69a3ebee94f51b4

```java
    public void testSplitKeyValueEscapedEquals()  {
        String test = "Truth\\=Beauty";
        String[] kv = Ini.Section.splitKeyValue(test);
        assertEquals("Truth", kv[0]);
    }
```

The above was existing behavior. Though the more I look at it this must have been a side effect/bug.  I cannot imagine why you would want this behavior, I would have expected `Truth\\=Beauty` to be evaluated to `Truth\=Beauty` with a `null` value, or possibly a runtime parsing error if the value is required.

I'm guessing this might have been the result of the key and value logic sharing the same escape logic.

TL;DR, I think this is a bug, and we should fix it in any version. (We do need to make this clear in the release notes, anyone caught by this _should_ be able to process the INI file/string to strip out the odd `\\=` if needed.

Thoughts?

bmarwell:
+1. Cherry pick into 1.7.x, remove the test. This behaviour can be restored easily by removing one backslash and does not look intentional.

fpapon:
Ok I will cherrypick to 1.7.x branch and add info in the release note.
Thanks for your feedback!

todmorrison:
This is a cleaner way and should fix up the "escaped escape-char" case (i.e. handle "//" as a single "/")

`                if (buildingKey) {
                  // Given that isCharEscaped actually means isEscapeChar, the middle case is always
                  // false.
                  // if (isKeyValueSeparatorChar(c) && !isCharEscaped(line, i) && !isCharEscaped(line, i-1)) {
                    if (isKeyValueSeparatorChar(c) &&  !isCharEscaped(line, i-1)) {
                        buildingKey = false;//now start building the value
                  // Allow escaped escape character -- adjusting for the misleading naming of isCharEscaped
                    } else if (!isCharEscaped(line, i) || !isCharEscaped(line, i-1) ){
                        keyBuffer.append(c);
                    }
`

fpapon:
@todmorrison thanks for your feedback, I will take a look.

fpapon:
@todmorrison I'm not sure to understand your code because our test failed. Did you take a look on our utests?

todmorrison:
Hi @fpapon, 

You're right. I partially amend my comment. The first part, changing:

```if (isKeyValueSeparatorChar(c) && !isCharEscaped(line, i) && !isCharEscaped(line, i-1)) ```

to 

```if (isKeyValueSeparatorChar(c) &&  !isCharEscaped(line, i-1))```

is correct. The middle case is redundant since `isKeyValueSeparatorChar(c)` where `c = line.charAt(i)` means the character at `i` is "=" and certainly not "\".

For the second part, `(!isCharEscaped(line, i) || !isCharEscaped(line, i-1) `, this fails the test for "Tru\\th=Beauty" = "Truth", which still doesn't look right to me but my "fix" doesn't solve the problem either. Specifically, I feel that backslashes should be allowed in the key (and possibly other escape sequences), but this is harder than I originally thought. (A project for another day).

Anyway, I do feel my observation that the middle case in the first conditional is redundant is worth fixing. I withdraw the second change though.

PR #288, opened 2021-03-21
Bugfix, fixed Scanner issue

pitjazz:

 - [ ] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/SHIRO) filed 
 - [ ] Format the pull request title like `[SHIRO-XXX] - Fixes bug in SessionManager`,
 - [ ] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
 - [ ] Run `mvn clean install apache-rat:check` to make sure basic checks pass. A more thorough check will be performed on your pull request automatically.
 
### DIFF
--- a/core/src/main/java/org/apache/shiro/realm/text/TextConfigurationRealm.java
+++ b/core/src/main/java/org/apache/shiro/realm/text/TextConfigurationRealm.java
@@ -211,9 +211,10 @@ public class TextConfigurationRealm extends SimpleAccountRealm {
 
     protected static Set<String> toLines(String s) {
         LinkedHashSet<String> set = new LinkedHashSet<String>();
-        Scanner scanner = new Scanner(s);
-        while (scanner.hasNextLine()) {
-            set.add(scanner.nextLine());
+        try (Scanner scanner = new Scanner(s)) {
+            while (scanner.hasNextLine()) {
+                set.add(scanner.nextLine());
+            }
         }
         return set;
     }


core/src/main/java/org/apache/shiro/realm/text/TextConfigurationRealm.java
```diff
@@ -211,11 +212,14 @@ protected void processUserDefinitions(Map<String, String> userDefs) {
...
-        }
-        return set;
+		try (Scanner scanner = new Scanner(s)) {
+			while (scanner.hasNextLine()) {
+				set.add(scanner.nextLine());
+			}
+		} catch (NoSuchElementException e) {
+			e.printStackTrace();
+		} 
+		return set;
```
bdemers, line 222: The rest of the file is formatted with spaces, this blog looks like tabs? can you fix that?

Instead of `e.printStackTrace()` use a logger.

More importantly though, how was this the exception triggered?  I'm guessing if `hasNextLine()` returns true, and then `.nextLine()` throws a `NoSuchElementException`, that's not an error you would want to recover from, as the stream would have been corrupted?  And in this case potentially worse because we are operating against a String in memory, which means things have gone severely wrong?

I could be miss understanding something here though, I only took a quick look.

bdemers (changes requested):
Thanks @pitjazz!

It's not clear what this PR fixes,  can you update the PR title with a specific fix,
And add detail to the commit/PR description

Adding a test would be great too!

core/src/main/java/org/apache/shiro/realm/text/TextConfigurationRealm.java
pitjazz: Link to Apache's SonarCloud: [the bug](https://sonarcloud.io/project/issues?id=apache_shiro&languages=java&open=AXTvO9kn3S_U1xqKcObL&resolved=false&severities=BLOCKER&types=BUG).

bmarwell: > how was this the exception triggered? 

> Link to Apache's SonarCloud: the bug.

yields: `Use try-with-resources or close this "Scanner" in a "finally" clause.`

That is, you would not want to catch the exception. Just the try would be fine in this case, as this is only about closing the scanner. But as we are reading a String, this is a noop anyway:

https://github.com/openjdk/jdk/blob/d7268fa3a68c2356548651a27b307d7f7158e700/src/java.base/share/classes/java/util/Scanner.java#L1171-L1175

So in this case, while Sonar is right in the sense of "good habit" (as the source could change): My -1 on this as we are not "fixing" anything or making anything better. Closing the scanner is fine for me, but there was no underlying bug and the API would also never be able to change to something `Closeable`.

core/src/main/java/org/apache/shiro/realm/text/TextConfigurationRealm.java
```diff
@@ -211,9 +215,15 @@ protected void processUserDefinitions(Map<String, String> userDefs) {
...
-            set.add(scanner.nextLine());
+        try (Scanner scanner = new Scanner(s)) {
+            while (scanner.hasNextLine()) {
+                set.add(scanner.nextLine());
+            }
+        } catch (Exception e) {
+            if (log.isWarnEnabled()) {
+                String msg = "Unable to fetch next line, Scanner stream corrupted.";
+                log.warn(msg, e);
+            }
```
bmarwell, line 226: Do not catch a generic exception. In this case, do not catch anything at all. We would want the Exception to promote here, otherwise the behaviour would change.

core/src/test/java/org/apache/shiro/realm/text/TextConfigurationRealmTest.java
```diff
@@ -247,6 +249,21 @@ public void run() {
...
+     * Test that scanner reads next line
+     */
+    @Test
+    public void testScannerHasNextLine() {
+        realm = new TestRealm() {
+            public void test(Thread runnable) throws InterruptedException {
+                Scanner scanner = new Scanner("Scanner\nTest\n");
+                Assert.assertNotNull(scanner);
+                assertEquals("Scanner", scanner.nextLine());
+                assertEquals("Test", scanner.nextLine());
```
bmarwell, line 263: You are testing `java.util.Scanner`, not `TextConfigurationRealm` you modified.

bmarwell (changes requested):

core/src/main/java/org/apache/shiro/realm/text/TextConfigurationRealm.java
bdemers: Agree, we want the original exception thrown here, failure to parse the config (especially config related to security) is a valid reason to throw an exception and resulting in a failure of the application to start.

A couple of other nits, as you learn more about Java:

you don't need to check `log.isWarnEnabled()` unless you are concatenating strings, or doing another expensive operation in that block, the log back `warn` method will do this check for you. 

But like we mentioned, we don't want to swallow this exception

@pitjazz let's close this PR for now, if there is a listing issue you want to fix, we can continue the thread here and provide other suggestions if needed

bdemers, line 226:
```suggestion
        }
```

I'm guessing just removing the catch block should remove the sonar error.

bdemers:
@pitjazz actually, reopened the PR with a suggestion, as I _think_ that is what you are trying to resolve

core/src/main/java/org/apache/shiro/realm/text/TextConfigurationRealm.java
pitjazz: I removed the catch block, updated the pull request. What I want to do is just close the scanner and catch is not needed for that.

I also removed the unit test. There is no need to test how scanner reads the nextLine, because reading of users and roles are already tested in TextConfigurationRealmTest. Perhaps closing the scanner could be tested, but IMHO it's not relevant.

bdemers (changes requested):
Code looks good now!

Can you fix the commit message (some of it is no longer valid)

pitjazz:
> Code looks good now!
> 
> Can you fix the commit message (some of it is no longer valid)

Commit message fixed.

bdemers:
Thanks @pitjazz

PR #297, opened 2021-05-04
1.8 build issues

bdemers:
Fix version and build issues on 1.8 branch

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -80,7 +80,7 @@ pipeline {
         stage('Tests') {
             steps {
                 echo 'Running tests'
-                sh 'mvn verify -Prun-its'
+                sh 'mvn verify -Prun-its ${MVN_LOCAL_REPO_OPT} -Dinvoker.streamLogsOnFailures=true'
             }
             post {
                 always {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -69,10 +69,16 @@ pipeline {
             }
         }
 
+        stage('Apache Release Audit Tool') {
+            steps {
+                echo 'Rat'
+                sh 'mvn -U -B -e apache-rat:check'
+            }
+        }
+
         stage('Build') {
             steps {
                 echo 'Building'
-                sh 'mvn -U -B -e apache-rat:check'
                 sh 'mvn -U -B -e ${MVN_LOCAL_REPO_OPT} clean install -DskipTests'
             }
         }
@@ -93,7 +99,7 @@ pipeline {
         stage('Generate doc') {
             when {
                 expression {
-                    env.BRANCH_NAME ==~ /(1.6.x|1.7.x|master)/
+                    env.BRANCH_NAME ==~ /(1.7.x|1.8.x|main)/
                 }
             }
             steps {
@@ -105,7 +111,7 @@ pipeline {
         stage('Deploy') {
             when {
                 expression {
-                    env.BRANCH_NAME ==~ /(1.6.x|1.7.x|master)/
+                    env.BRANCH_NAME ==~ /(1.7.x|1.8.x|main)/
                 }
             }
             steps {
@@ -120,7 +126,7 @@ pipeline {
         // If this build failed, send an email to the list.
         failure {
             script {
-                if(env.BRANCH_NAME == "1.6.x" || env.BRANCH_NAME == "1.7.x" || env.BRANCH_NAME == "master") {
+                if(env.BRANCH_NAME == "1.7.x" || env.BRANCH_NAME == "1.8.x" || env.BRANCH_NAME == "main") {
                     emailext(
                             subject: "[BUILD-FAILURE]: Job '${env.JOB_NAME} [${env.BRANCH_NAME}] [${env.BUILD_NUMBER}]'",
                             body: """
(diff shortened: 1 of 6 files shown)


crypto/cipher/src/main/java/org/apache/shiro/crypto/AesCipherService.java
```diff
@@ -89,12 +89,16 @@
      * <b>**</b>Since {@code GCM} is a stream cipher, padding is implemented in the operation mode and an external padding scheme
      * cannot be used in conjunction with {@code GCM}. In fact, {@code AES/GCM/PKCS5Padding} is just an alias in most JVM for
      * {@code AES/GCM/NoPadding}.
+     * <p/>
+     * <b>NOTE:</b> As of Java 14, setting a streaming padding for the above example will throw a NoSuchAlgorithmException
+     * @see <a href="https://www.oracle.com/java/technologies/javase/14-relnote-issues.html#JDK-8180392">JDK-8180392</a>
```
bdemers, line 94: I'll merge this line back to master

pom.xml
```diff
@@ -110,7 +110,7 @@
         <!-- Test 3rd-party dependencies: -->
         <easymock.version>4.1</easymock.version>
         <gmaven.version>1.8.0</gmaven.version>
-        <groovy.version>2.5.8</groovy.version>
+        <groovy.version>2.5.14</groovy.version>
```
bdemers, line 113: newer groovy uses an ASM version that supports Java 14

web/src/it/projects/MultipleServiceLoader/pom.xml
```diff
@@ -38,7 +38,7 @@
         <dependency>
             <groupId>org.apache.shiro</groupId>
             <artifactId>shiro-web</artifactId>
-            <version>1.7.1-SNAPSHOT</version>
+            <version>@project.version@</version>
```
bdemers, line 41: I'll double-check `main` to make sure this is also in place

Jenkinsfile-jdk11
```diff
@@ -72,14 +72,15 @@ pipeline {
         stage('Build') {
             steps {
                 echo 'Building'
-                sh 'mvn -U -B -e clean install -DskipTests apache-rat:check'
+                sh 'mvn -U -B -e apache-rat:check'
```
bdemers, line 75: A bit of a cheat, execute "rat" using the global repo (shared with all builds) `~/.m2/repository`
Other maven executions use `.repository` in the clean checkout directory.

Alternatively, I can configure the rat plugin to ignore `.repository` and add it to .gitignore?

pom.xml
```diff
@@ -1533,6 +1533,7 @@
                     <plugin>
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-invoker-plugin</artifactId>
+                        <version>3.2.2</version>
```
bdemers, line 1536: starting with this version `streamLogsOnFailures` has been added (to help with logging output for CI usage)

Jenkinsfile
```diff
@@ -72,14 +72,15 @@ pipeline {
...
+                sh 'mvn -U -B -e apache-rat:check'
+                sh 'mvn -U -B -e ${MVN_LOCAL_REPO_OPT} clean install -DskipTests'
             }
         }
 
         stage('Tests') {
             steps {
                 echo 'Running tests'
-                sh 'mvn verify -Prun-its'
+                sh 'mvn verify -Prun-its ${MVN_LOCAL_REPO_OPT} -Dinvoker.streamLogsOnFailures=true'
```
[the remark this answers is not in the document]
bmarwell: Yes, because you don't want them on your PC, probably. You have build.log files there. That's what I find much more comfortable than scrolling through a console or even copy lots from there.

Jenkinsfile-jdk11
bmarwell: Sigh. That's why maven has a per-workspace repo. That shouldn't be needed and became a little better through the matrix build on main.

Or just do the rat check on jdk11 only.

Jenkinsfile-jdk11
```diff
@@ -72,14 +72,15 @@ pipeline {
...
+                sh 'mvn -U -B -e ${MVN_LOCAL_REPO_OPT} clean install -DskipTests'
```
bmarwell, line 76: Is install really needed?

pom.xml
bmarwell: Good option!

Jenkinsfile-jdk11
bdemers: just verify would be fine, but I'm not sure why the build and test were broken into different stages originally?

Maybe the stages should be:
* apache-rat:check
* dependency:resolve dependency:plugin-resolve
* verify -Prun-its

Jenkinsfile-jdk11
bdemers: +1 for this, but I'm going to move it in its own stage for now, and try to remove the duplication between these files on `main`.

Hopefully something like this in each Jenkins file:
```groovy
const jdkVersion = 'jdk_8_latest'
const deploy = false
const rat = false
const supportedBranch = (env.BRANCH_NAME == "1.7.x" || env.BRANCH_NAME == "1.8.x" || env.BRANCH_NAME == "main") // we have a lot of this in the individual files

load('Jenkinsfile-common')
}
```

I'm not sure that will work (given the scope of those globals 🤮), but I'll poke around with it

bdemers: oh, this is already resolved on `main` with a matrix build 😄

PR #298, opened 2021-05-06
Pr 297 to main

bdemers:


.jenkins.groovy
```diff
@@ -17,6 +17,8 @@
  * under the License.
  */
 
+def supportedBranch = env.BRANCH_NAME ==~ /(1.7.x|1.8.x|main)/
```
bmarwell, line 20: "supported"? Like in "we give support"? Maybe just change to "deployable branch" or "maintained" or something like this. It should be clear that this *only* has to do with snapshot deployments.

bmarwell (changes requested):

.jenkins.groovy
bdemers: Good call, I wasn't happy with this name either, `deployableBranch` sounds good to me!

bdemers: @bmarwell back to you when you get a chance

bmarwell (approved):

PR #299, opened 2021-05-07
jenkins matrix 1.8

bdemers:


.jenkins.groovy
```diff
@@ -0,0 +1,186 @@
...
+    stages {
+        stage('Build') {
+            matrix {
+                axes {
+                    axis {
+                        // https://cwiki.apache.org/confluence/display/INFRA/JDK+Installation+Matrix
+                        name 'MATRIX_JDK'
+                        values 'jdk_1.8_latest',
+                                'jdk_11_latest',
+                                'jdk_14_latest'
```
bmarwell, line 41: Maybe add the other jdks as we have on main? Because OoenJ9 had given us headaches in the past.

.jenkins.groovy
```diff
@@ -0,0 +1,186 @@
...
+                        steps {
+                            echo 'License check'
+                            sh 'mvn --batch-mode -Drat.consoleOutput=true apache-rat:check'
+                        }
+                    }
+
+                    stage('Build') {
+                        steps {
+                            echo 'Building'
+                            sh 'mvn --update-snapshots --batch-mode --errors clean verify -Pdocs -Dmaven.test.failure.ignore=true'
```
bmarwell, line 96: We shouldn't be using snapshots, so updating should not be needed.

bmarwell (changes requested):

.jenkins.groovy
bdemers: I don't disagree, but I'm trying to get the build stable first, I think there is a flakey test in here somewhere, but... it could have been something i tweaked too

.jenkins.groovy
bdemers: Fixed :)

.jenkins.groovy
fpapon: Latest JDK is 16 but I'm not sure we need to support non-LTS JDK

fpapon (approved):

.jenkins.groovy
bmarwell: Yes, this is just a build. It's not about support. We should test on as many JDKs as possible to find about problems with the next LTS as early as possible. So, -14, +16

bmarwell: @bdemers the test I removed from main?

bmarwell (approved):

bmarwell (changes requested):

.jenkins.groovy
```diff
@@ -0,0 +1,186 @@
...
+    stages {
+        stage('Build') {
+            matrix {
+                axes {
+                    axis {
+                        // https://cwiki.apache.org/confluence/display/INFRA/JDK+Installation+Matrix
+                        name 'MATRIX_JDK'
+                        values 'jdk_1.8_latest',
+                                'jdk_11_latest',
+                                'jdk_16_latest'
```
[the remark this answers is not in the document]
bdemers: The maven bundle plugin doesn't build with 16, and there are some groovy compatibility issues too.

I think it's worth building up to Java 14 (there was a crypto alias change) for the time being on this branch, and keep `main` building on 8, 11, and the last release (16 atm).

Thoughts?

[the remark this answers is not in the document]
bmarwell: Ok, I think we can keep 14 for now.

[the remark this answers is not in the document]
fpapon: +1

.jenkins.groovy
bdemers: It looked like a Groovy + ASM problem, so I didn't spend any more time with it

fpapon: @bdemers +1 to move forward ;)

bmarwell (approved):

PR #311, opened 2021-07-16
use block list to block dangerous gadget

k4n5ha0:
org.apache.commons.collections.functors.ChainedTransformer.transform
org.apache.commons.collections.functors.InvokerTransformer
org.apache.commons.collections.functors.InstantiateTransformer
org.apache.commons.collections4.functors.InvokerTransformer
org.apache.commons.collections4.functors.InstantiateTransformer
com.sun.org.apache.xalan.internal.xsltc.trax.TemplatesImpl
org.apache.xalan.xsltc.trax.TemplatesImpl
java.util.PriorityQueue

these gadget is  dangerous  attack gadget.if we block these necessary class. 
we can reduce vul.
thx

### DIFF
--- a/lang/src/main/java/org/apache/shiro/lang/io/ClassResolvingObjectInputStream.java
+++ b/lang/src/main/java/org/apache/shiro/lang/io/ClassResolvingObjectInputStream.java
@@ -39,9 +39,9 @@ public class ClassResolvingObjectInputStream extends ObjectInputStream {
     }
 
     /**
-     * Resolves an {@link ObjectStreamClass} by delegating to Shiro's 
+     * Resolves an {@link ObjectStreamClass} by delegating to Shiro's
      * {@link ClassUtils#forName(String)} utility method, which is known to work in all ClassLoader environments.
-     * 
+     *
      * @param osc the ObjectStreamClass to resolve the class name.
      * @return the discovered class
      * @throws IOException never - declaration retained for subclass consistency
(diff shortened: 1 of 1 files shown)


bdemers:
Thanks for reaching out @k4n5ha0!  I want to point out a few things first.

Potential vulnerabilities need to be handled differently than regular software bugs:
- Apache has a [great guide on the topic](https://www.apache.org/security/), and a [step-by-step guide for committers](https://www.apache.org/security/committers.html).
- I have a [few posts](https://developer.okta.com/blog/2020/03/13/developers-guide-on-reporting-vulnerabilities) on the topic as well.

Back to the topic:
- Shiro makes use of this ObjectStream through an encrypted stream, this means that Shiro itself should be the only one serializing the original data (mitigating this risk).
- This PR implements a block list, which doesn't scale well, there will always be other entries needed to add to a block list.  instead, an allow list would be recommended, especially for uses in security. (e.g. only allow classes `A`, `B`, and `C`, instead of only blocking `Z`, `Y`, and `Z`)


All that said, we can continue the thread on the Shiro Security list if you want (or if you see any gaps in my explaination)!
security@shiro.apache.org

k4n5ha0:
@bdemers  pls show some example ， how to override ```org.apache.shiro.io.DefaultSerializer``` i get a idea
thx :-)

bdemers:
@k4n5ha0, it depends on how you have your application configured, but if you are using a `shiro.ini` file (or similar) it should be something like:

```ini
# define your implementation
myCustomSerializer = com.example.MySerializer

# assign it
securityManager.rememberMeManager.serializer = $myCustomSerializer
```

There are a few other `ini` examples in Shiro's configuration docs: https://shiro.apache.org/configuration.html

PR #312, opened 2021-07-16
Add note about reporting a vuln in the PR template

bdemers:
NOTE, this is a comment in the template, which will be shown when a user creates a new PR, but NOT shown in the rendered text.


bmarwell (approved):
Yes, please!

PR #318, opened 2021-08-18
Add @Documented to Shiro annotations

bdemers:
Shiro annotations will now be included in generated Javadoc

Fixes: SHIRO-216


bmarwell (approved):
Question: No sure about how much we want to document, but there is also:
`org.apache.shiro.event.Subscribe`.
Do we need to take care of this annotation as well?

bdemers:
I think you are right, given that this is an entry point into a code path (when an event is triggered), adding `@Documented` should help make it more obvious.  I'll push a change in a few minutes 😄

bdemers:
(amended previous commit)

PR #319, opened 2021-08-30
SonarCloud issues fixed

sp-maintenance-course:
SonarCloud issues fixed for the samples/guice/src/main/webapp/login.jsp

### DIFF
--- a/samples/guice/src/main/webapp/login.jsp
+++ b/samples/guice/src/main/webapp/login.jsp
@@ -18,8 +18,11 @@
   --%>
 <%@ include file="include.jsp" %>
 
-<html>
+<!DOCTYPE html>
+<html lang="en">
+
 <head>
+	<title>Webapp sample login page</title>
     <link type="text/css" rel="stylesheet" href="<c:url value="/style.css"/>"/>
 </head>
 <body>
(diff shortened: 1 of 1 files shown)


bdemers (changes requested):
Thanks, @sp-maintenance-course !

The changes look good but can you tweak the commit and PR description to explain what and why this change was made?

Referencing SonarCloud doesn't help for folks how have never seen the report.

Maybe something like:

```txt
Updating sample HTML file to use best practices

Issues discovered via SonarCloud: https://link-if/possible
* Added HTML missing language
* Added page title
* Replaced deprecated HTML tags with css equivalent
```

(I'm guessing SonarCloud probably has some descriptions of the issues you could use too, my comment above are just guesses based on the commit)

(I have one other question inline)

samples/guice/src/main/webapp/login.jsp
```diff
@@ -58,11 +61,12 @@
...
 
-    <table class="sample">
+    <table class="sample" style="padding: 10px;">
+    	<caption style="padding: 10px;"> Sample accounts </caption>
         <thead>
         <tr>
-            <th>Username</th>
-            <th>Password</th>
+            <th id="username">Username</th>
+            <th id="password">Password</th>
```
bdemers, line 69: Can you explain this change?  I'm guessing if `id` attributes are set they should probably be set on the actual inputs?

I don't have the context of why this change is suggested, by my assumption adding the `id` attr base it easier to drive test frameworks like Selenium (so a `getById()` can be used instead of a more complicated selector?)

PR #350, opened 2022-03-18
[SHIRO-871] Apply principalSuffix only when username does NOT already contain the suffix

bdemers:
Fixes: SHIRO-871


manticore-projects:
Thanks Brian,

unfortunately the patch needs a fix:
```java
if (principalSuffix != null && !userPrincipalName.toLowerCase().endsWith( principalSuffix.toLowerCase() ) ) {
```

Apologies for working sloppy.

bdemers:
Thanks @manticore-projects! I caught that one, take a look at the PR, I tweaked the original patch a little and added a couple of tests. Any feedback welcome!

PR #362, opened 2022-06-17
Reduce the scope of the ThreadLocals used in SecurityManagerTestSupport

bdemers:
Individual test methods can scope the usage of a SecurityManger and Subject to specific methods.

Related to: #361


bmarwell (approved):
Nice! Thanks! 👍🏻

PR #365, opened 2022-06-29
[SHIRO-884] use plugin version from parent

hboutemy:
I just removed the plugin version to simply inherit from parent plugin management

but I don't see why this doc profile is used given the work is already done by the apache-release profile: another fix would basically to fully remove this profile


fpapon (approved):
LGTM
@hboutemy thanks!

PR #372, opened 2022-09-05
[SHIRO-512] catch SessionException in getRunAsPrincipalsStack

lprimak:
Fixes [SHIRO-512](https://issues.apache.org/jira/browse/SHIRO-512)
Fixes https://github.com/flowlogix/flowlogix/issues/284

 - [X] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/SHIRO) filed 
 - [X] Each commit in the pull request should have a meaningful subject line and body.
 - [X] Format the pull request title like `[SHIRO-XXX] - Fixes bug in SessionManager`,
 - [X] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
 - [X] Run `mvn clean install apache-rat:check` to make sure basic checks pass. A more thorough check will be performed on your pull request automatically.
 - [X] If you have a group of commits related to the same change, please squash your commits into one and force push your branch using `git rebase -i`. 
 
 - [X] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
 - [X] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

### DIFF
--- a/core/src/main/java/org/apache/shiro/subject/support/DelegatingSubject.java
+++ b/core/src/main/java/org/apache/shiro/subject/support/DelegatingSubject.java
@@ -474,6 +474,10 @@ public class DelegatingSubject implements Subject {
             try {
                 return (List<PrincipalCollection>) session.getAttribute(RUN_AS_PRINCIPALS_SESSION_KEY);
             } catch (SessionException se) {
+                // There could be a rare race condition when a session is invalidated in another thread,
+                // this thread could throw this exception, so we catch it
+                // similar issue as in clearRunAsIdentitiesInternal()
+                // See https://issues.apache.org/jira/browse/SHIRO-512
                 log.debug("Encountered session exception trying to get 'runAs' principal stack.  This "
                         + "can generally safely be ignored.", se);
             }
--- a/core/src/main/java/org/apache/shiro/subject/support/DelegatingSubject.java
+++ b/core/src/main/java/org/apache/shiro/subject/support/DelegatingSubject.java
@@ -471,7 +471,12 @@ public class DelegatingSubject implements Subject {
     private List<PrincipalCollection> getRunAsPrincipalsStack() {
         Session session = getSession(false);
         if (session != null) {
-            return (List<PrincipalCollection>) session.getAttribute(RUN_AS_PRINCIPALS_SESSION_KEY);
+            try {
+                return (List<PrincipalCollection>) session.getAttribute(RUN_AS_PRINCIPALS_SESSION_KEY);
+            } catch (SessionException se) {
+                log.debug("Encountered session exception trying to get 'runAs' principal stack.  This "
+                        + "can generally safely be ignored.", se);
+            }
         }
         return null;
     }


sunshineandy:
这是自动回复邮件。你好，您的邮件已经发送到我的邮箱，我看过后会尽快给您回复。

lprimak:
Something's wrong with Jenkins environment:
```
com.github.mjeanroy.junit.servers.exceptions.ServerStartException: java.io.IOException: Failed to bind to 0.0.0.0/0.0.0.0:37081
Caused by: java.io.IOException: Failed to bind to 0.0.0.0/0.0.0.0:37081
Caused by: java.net.BindException: Address already in use
```

fpapon (approved):
LGTM
Thanks for your contribution!

core/src/main/java/org/apache/shiro/subject/support/DelegatingSubject.java
```diff
@@ -471,7 +471,12 @@ public PrincipalCollection releaseRunAs() {
     private List<PrincipalCollection> getRunAsPrincipalsStack() {
         Session session = getSession(false);
         if (session != null) {
-            return (List<PrincipalCollection>) session.getAttribute(RUN_AS_PRINCIPALS_SESSION_KEY);
+            try {
+                return (List<PrincipalCollection>) session.getAttribute(RUN_AS_PRINCIPALS_SESSION_KEY);
+            } catch (SessionException se) {
+                log.debug("Encountered session exception trying to get 'runAs' principal stack.  This "
+                        + "can generally safely be ignored.", se);
+            }
```
bmarwell, line 479: Can you add some more information? Not sure whether it should be a comment or go into the exception message, though. Can you elaboare, WHY this can be ignored and WHY and WHEN this occurs? Is it true it only occurs on Tomcat? Shouldn't we file an issue there instead?

fpapon: This exception seems to be a race condition on Tomcat but appear randomly so I'm not sure we can find more information in the exception.

rzo1: The underlying `IllegalStateException` is thrown, if the session is invalidated (servlet spec).

lprimak: Please see the underlying JIRA. It's not only Tomcat, but any other servlet container.
I have reproduced it on both Payara, Glassfish and Jetty.
There is also the same try/catch block on `clearRunAsIdentitiesInternal()` for the same reason.

I did add some comments to the code

bdemers (approved):

core/src/main/java/org/apache/shiro/subject/support/DelegatingSubject.java
```diff
@@ -471,7 +471,16 @@ public PrincipalCollection releaseRunAs() {
...
         if (session != null) {
-            return (List<PrincipalCollection>) session.getAttribute(RUN_AS_PRINCIPALS_SESSION_KEY);
+            try {
+                return (List<PrincipalCollection>) session.getAttribute(RUN_AS_PRINCIPALS_SESSION_KEY);
+            } catch (SessionException se) {
+                // There could be a rare race condition when a session is invalidated in another thread,
+                // this thread could throw this exception, so we catch it
+                // similar issue as in clearRunAsIdentitiesInternal()
+                // See https://issues.apache.org/jira/browse/SHIRO-512
+                log.debug("Encountered session exception trying to get 'runAs' principal stack.  This "
```
adamenveil, line 481: I'm not trying to be rude, I'm legitimately trying to understand, how is this safe to ignore? This appears to be the reason one of my unit tests is failing now after bumping Shiro up to `1.10.0`

So if Thread_1 invalidates the subject, then Thread_2 attempts to get the session, this block here would originally throw an exception to Thread_2 informing that the session has been invalidated. But now Thread_2 will catch and just log the exception instead. This would essentially let Thread_2 believe the subject is still authenticated, when the session should actually be invalidated.

bdemers: @adamenveil 
I can sort of see both sides of this.  I think what this change was trying to prevent was:

1. Thread 1 - calls `Session session = getSession(false);` // valid session returned
1. Thread 2 - calls `session.invalidate()`; // somewhere else in the code
1. Thread 1 - calls `session.getAttribute(RUN_AS_PRINCIPALS_SESSION_KEY);` // that throws an exception, the session is invalid.
1. Thread 1 - the `getRunAsPrincipalsStack` method returns null

If this same flow of events happened in a different order:
1. Thread 1 - calls `Session session = getSession(false);` // valid session returned
1. Thread 1 - calls `session.getAttribute(RUN_AS_PRINCIPALS_SESSION_KEY);` // object returned
1. Thread 2 - calls `session.invalidate()`; // somewhere else in the code

The caller of `getRunAsPrincipalsStack()` doesn't know the session has been invalidated.

Or in the order of:

1. Thread 2 - calls `session.invalidate()`; // somewhere else in the code
1. Thread 1 - calls `Session session = getSession(false);` // null is returned
1. Thread 1 - the `getRunAsPrincipalsStack` method returns null


Can you give us an idea of what your code is doing maybe there is something else we could do?

lprimak: i.e. it's not specified that the subject is automatically logged out when session is invalidated.
Proper way (according to the API) to log a subject out is `SecurityUtils.getSubject().invalidate()` and not `session.invalidate()`

It is true that Shiro *could* add HttpSession listeners and automatically log the subject out if a session gets invalidated, but neither the old nor new code does this, perhaps this could be an additional feature in the future, but certainly not a showstopper

PR #410, opened 2022-10-05
Update OWASP dep check plugin and Karaf versions

bdemers:
- Update owasp dependency check
- Updating Karaf version


fpapon (approved):
Tested with Apache Karaf 4.4.1

PR #518, opened 2022-10-28
Allow for direct configuration of ShiroFilter through WebEnvironment

bdemers:
Adds new ShiroFilterConfiguration class exposed through the WebEnvironment
This class allows for confiuration of the Shiro filter through various config mechanisms (Ini, Guice, Spring, etc)
This makes it easer to enable a static security manager when using Shiro's Servlet fragment, as now a web.xml is not required

Backport: 28e10e0ca1cdcd2cede86802fde8464b29265fc8

### DIFF
--- a/support/guice/src/test/java/org/apache/shiro/guice/web/GuiceShiroFilterTest.java
+++ b/support/guice/src/test/java/org/apache/shiro/guice/web/GuiceShiroFilterTest.java
@@ -19,13 +19,13 @@
 package org.apache.shiro.guice.web;
 
 import com.google.inject.spi.InjectionPoint;
+import org.apache.shiro.web.config.ShiroFilterConfiguration;
 import org.apache.shiro.web.filter.mgt.FilterChainResolver;
 import org.apache.shiro.web.mgt.WebSecurityManager;
 import org.junit.Test;
 
-import static org.easymock.EasyMock.createMock;
-import static org.junit.Assert.assertSame;
-import static org.junit.Assert.fail;
+import static org.easymock.EasyMock.*;
+import static org.junit.Assert.*;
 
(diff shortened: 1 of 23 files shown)


bdemers:
This should wait for #519 to be merged.

support/guice/src/test/java/org/apache/shiro/guice/web/GuiceShiroFilterTest.java
```diff
@@ -19,13 +19,13 @@
...
 import com.google.inject.spi.InjectionPoint;
+import org.apache.shiro.web.config.ShiroFilterConfiguration;
 import org.apache.shiro.web.filter.mgt.FilterChainResolver;
 import org.apache.shiro.web.mgt.WebSecurityManager;
 import org.junit.Test;
 
-import static org.easymock.EasyMock.createMock;
-import static org.junit.Assert.assertSame;
-import static org.junit.Assert.fail;
+import static org.easymock.EasyMock.*;
```
bmarwell, line 27: Please revert to explicit imports.

web/src/main/java/org/apache/shiro/web/servlet/ShiroFilter.java
```diff
@@ -70,8 +70,10 @@ public class ShiroFilter extends AbstractShiroFilter {
      */
     @Override
     public void init() throws Exception {
+
```
bmarwell, line 73: ?

bmarwell (commented):
Thanks! You even rewrote the whole test, which is great! Just a nit about star imports, I'd like to keep the former code style (so it's consistent). Maybe we should discuss light auto-formatting.

PR #522, opened 2022-11-02
[SHIRO-889] Provide Jakarta jar modules

fpapon:


bmarwell:
Maybe we can user undertow?
Here's a basic setup without deployment and empty instance manager.
All you need to do: configure the `.war` and stop the server again later.

Switched to non-TLS for test purposes.

https://github.com/fpapon/shiro/compare/SHIRO-889...apache:shiro:SHIRO-889_undertow

WDYT?

fpapon:
> Maybe we can user undertow? Here's a basic setup without deployment and empty instance manager. All you need to do: configure the `.war` and stop the server again later.
> 
> Switched to non-TLS for test purposes.
> 
> [fpapon/shiro@SHIRO-889...apache:shiro:SHIRO-889_undertow](https://github.com/fpapon/shiro/compare/SHIRO-889...apache:shiro:SHIRO-889_undertow)
> 
> WDYT?

Thank you! I will take a look :)

samples/web-jakarta/src/test/java/org/apache/shiro/test/web/jakarta/JakartaAbstractContainerIT.java
```diff
@@ -0,0 +1,120 @@
...
+    @AfterClass
+    public static void stopContainer() {
+        if (meecrowave != null) {
+            meecrowave.close();
+        }
+    }
+
+    // Dealing with a keystore is NOT fun, it's easier to script one with the keytool
+    // see src/main/resources/createKeyStore.sh for more info
+    private static File setupKeyStore() {
```
bmarwell, line 105: This can be removed, as this test now uses http which is sufficient

samples/web-jakarta/src/test/java/org/apache/shiro/test/web/jakarta/WebContainerTest.java
```diff
@@ -0,0 +1,69 @@
...
+                .request(TEXT_HTML_TYPE)
+                .get(String.class);
+        HtmlPage page = webClient.getPage(getBaseUri() + "login.jsp");
+        HtmlForm form = page.getFormByName("loginform");
+        form.getInputByName("username").setValueAttribute("root");
+        form.getInputByName("password").setValueAttribute("secret");
+        page = form.getInputByName("submit").click();
+        // This'll throw an exception if not logged in
+        // todo fix it
+        //page.getAnchorByHref("/logout");
```
bmarwell, line 67: Don't include this please

samples/web-jakarta/src/test/java/org/apache/shiro/test/web/jakarta/WebContainerTest.java
```diff
@@ -0,0 +1,69 @@
...
+    protected final WebClient webClient = new WebClient();
+
+    @Before
+    public void startup() throws IOException {
+
+        // Make sure we are logged out
+        final HtmlPage homePage = webClient.getPage(getBaseUri());
+        try {
+            // todo fix it
+            //homePage.getAnchorByHref("/logout").click();
```
bmarwell, line 46: Log out or remove

samples/web-jakarta/src/test/resources/createKeyStore.sh
```diff
@@ -0,0 +1,65 @@
...
+# The generated certificate is good for 10 years, and should NOT need to be recreated until then (unless
+# changes to the certificate are needed).
+# Last run with Java 1.8.0_252
+#
+# Usage: For JVM based applications, the resulting keystore MUST be configured to in order for clients to accept TLS
+# connections.  Typical usage requires setting of the Java system property 'javax.net.ssl.trustStore' to the file path
+# of the keystore. Either by adding a command line parameter: `-Djavax.net.ssl.trustStore=/path/to/keystore` or
+# programmatically: `System.setProperty("javax.net.ssl.trustStore", "/path/to/keystore")`
+
+dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
```
bmarwell, line 32: This entire script can be removed, http

samples/web-jakarta/src/test/resources/test-keystore.pem
```diff
@@ -0,0 +1,22 @@
+-----BEGIN CERTIFICATE-----
```
bmarwell, line 1: Can be removed

bmarwell (changes requested):

lprimak:
Question: why is there separate packages to be renamed in the shade plugin ?
Can’t it just be done in a single javax->jakarta rename and not individual packages such as JaxRS etc?

samples/web-jakarta/src/test/java/org/apache/shiro/test/web/jakarta/WebContainerTest.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+package org.apache.shiro.test.web.jakarta;
+
+import jakarta.ws.rs.client.Client;
+import jakarta.ws.rs.client.ClientBuilder;
+import jakarta.ws.rs.client.Entity;
+import jakarta.ws.rs.core.Cookie;
+import jakarta.ws.rs.core.Response;
+import org.junit.Test;
+import org.junit.jupiter.api.Assertions;
```
bmarwell, line 27: You are mixing junit4 and junit-jupiter assertions here. Please completely switch to -jupiter for all new tests.

samples/web-jakarta/src/test/java/org/apache/shiro/test/web/jakarta/WebContainerTest.java
```diff
@@ -0,0 +1,64 @@
...
+import static jakarta.ws.rs.core.MediaType.TEXT_HTML_TYPE;
+
+public class WebContainerTest extends JakartaAbstractContainerIT {
+
+    @Test
+    public void logIn() {
+        final Client client = ClientBuilder.newClient();
+
+        try {
+            final Response loginPage = client.target(getBaseUri())
```
bmarwell, line 39: response needs to be `close()`d.  However, you might be able to use auto-closeable. https://jakarta.ee/specifications/platform/8/apidocs/javax/ws/rs/core/response#close--

samples/web-jakarta/src/test/java/org/apache/shiro/test/web/jakarta/WebContainerTest.java
```diff
@@ -0,0 +1,64 @@
...
+                    .path("/login.jsp")
+                    .request(TEXT_HTML_TYPE)
+                    .get();
+            final Cookie jsessionid = new Cookie("JSESSIONID",loginPage.getMetadata().get("Set-Cookie").get(0).toString().split(";")[0].split("=")[1]);
+            Assertions.assertTrue(loginPage.readEntity(String.class).contains("loginform"));
+
+            final Response loginAction = client.target(getBaseUri())
```
bmarwell, line 46: Same. Needs `close()`.

samples/web-jakarta/src/main/webapp/style.css
```diff
@@ -0,0 +1,47 @@
...
+
+#contentBox {
+    text-align: center;
+    width: 50%;
+    margin: auto;
+    margin-top: 50px;
+    color: black;
+    background: #eee;
+    border: thick black solid;
+}
```
bmarwell, line 47: line feed missing at end of file

samples/web-jakarta/src/main/webapp/index.jsp
```diff
@@ -0,0 +1,21 @@
...
+  ~ Unless required by applicable law or agreed to in writing,
+  ~ software distributed under the License is distributed on an
+  ~ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  ~ KIND, either express or implied.  See the License for the
+  ~ specific language governing permissions and limitations
+  ~ under the License.
+  --%>
+
+<%-- Forward the user to the home page --%>
+<jsp:forward page="home.jsp"/>
```
bmarwell, line 21: line feed missing at EOF

samples/web-jakarta/src/main/webapp/include.jsp
```diff
@@ -0,0 +1,22 @@
...
+  ~ software distributed under the License is distributed on an
+  ~ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  ~ KIND, either express or implied.  See the License for the
+  ~ specific language governing permissions and limitations
+  ~ under the License.
+  --%>
+<%@ page import="org.apache.shiro.SecurityUtils" %>
+<%@ taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core" %>
+<%@ taglib prefix="fmt" uri="http://java.sun.com/jsp/jstl/fmt" %>
+<%@ taglib prefix="shiro" uri="http://shiro.apache.org/tags" %>
```
bmarwell, line 22: LF at EOF

samples/web-jakarta/src/main/webapp/account/index.jsp
```diff
@@ -0,0 +1,36 @@
...
+<h2>Users only</h2>
+
+<p>You are currently logged in.</p>
+
+<p><a href="<c:url value="/home.jsp"/>">Return to the home page.</a></p>
+
+<p><a href="<c:url value="/logout"/>" onclick="document.getElementById('logout_form').submit();return false;">Log out.</a></p>
+<form id="logout_form" action="<c:url value="/logout"/>" method="post"></form>
+</body>
+</html>
```
bmarwell, line 36: +LF

samples/web-jakarta/src/test/java/org/apache/shiro/test/web/jakarta/WebContainerTest.java
fpapon: client is closed in finally so I don't think we need to close the response but I can close it also in finally.

samples/web-jakarta/src/main/webapp/style.css
fpapon: weird because I copy/paste the `web` module sample without modification...

samples/web-jakarta/src/test/java/org/apache/shiro/test/web/jakarta/WebContainerTest.java
fpapon: assertj is not needed :)

fpapon:
@bmarwell changes done

bdemers (approved):
I'm +0.5, I worry a little about the `jakarta` classifier, as the exclude * will drop all of the dependencies (so the user would probably need to add the other bits like shiro-core to their pom)
But... that's not the end of the world, and other projects are doing the same thing (e.g. meecrowave)
I think getting it in as is and getting feedback from interested parties makes it worth merging now.
(given other projects are doing something similar, I'm hesitant to try to suggest an over-engineered solution 🤣 :)

TL;DR - my opinion is to merge it

fpapon:
> I'm +0.5, I worry a little about the `jakarta` classifier, as the exclude * will drop all of the dependencies (so the user would probably need to add the other bits like shiro-core to their pom) But... that's not the end of the world, and other projects are doing the same thing (e.g. meecrowave) I think getting it in as is and getting feedback from interested parties makes it worth merging now. (given other projects are doing something similar, I'm hesitant to try to suggest an over-engineered solution rofl :)
> 
> TL;DR - my opinion is to merge it

Hi, yes but honestly I don't know how to do it differently without a lot of code...
I will add a page on the Shiro website to well document the usage of the jakarta classifier.

garydgregory:
Releasing takes time and effort, which will happen, just not right now.

bdemers:
>  without a lot of code...

@fpapon exactly my concern too! Let’s move forward with this PR 🚢

bmarwell (approved):

fpapon:
Thanks, now I will prepare a PR for 1.11.x

lucasdillmann:
Hello everyone.

First of all, a huge thanks on the work to provide the Jakarta JARs. 

Looking in the 2.0.0 nightly snapshots, I was able to find the artifacts over the shiro-web and shiro-spring modules, but I couldn't find them in the shiro-spring-boot, shiro-spring-boot-starter and shiro-spring-boot-web-started.

Isn't needed to create the jakarta variant JAR/POM on the Spring Boot modules as well?

lprimak:
@lucasdillmann I don't think Shiro spring is compatible with Spring 6 or SpringBoot 3 quite yet. Not sure if those modules can be correctly done. I am not a Spring user so I can't comment further.

lucasdillmann:
@lprimak Is there any way I could help you guys in order to check/test if is compatible?

For example, if I create a test project with some basic scenarios as a POC (using a Shiro fork just to change to the Jakarta APIs on the boot modules), will it be useful in any way?

lprimak:
If it's only that Jakarta relocations are needed, you can try to add the shade plugin in your fork the same way other modules are done. That would be very helpful and appreciated!
The issue is that I personally can't test it because I am not a spring user.
Thank you!

lucasdillmann:
I'll do it and get back here with the results soon. As far I've seen in the source code, the changes come down to one single import change [in this file](https://github.com/apache/shiro/blob/main/support/spring-boot/spring-boot-starter/src/main/java/org/apache/shiro/spring/config/web/autoconfigure/ShiroWebFilterConfiguration.java) (apart from the dependency graph between the modules).

lucasdillmann:
@lprimak I finished the tests here using Spring Boot 3.0.1, and everything seems to be working just fine.

The source code of the test project is available [here](https://github.com/lucasdillmann/shiro-spring-boot-test). I didn't uplod the changes in the Shiro source, but it is just the inclusion of the maven-shade-plugin in the shiro-spring-boot module.

In short, these are the (exploratory) tests I've made:
- Check if subject is authenticated or not using both the annotations (RequiresUser, RequiresGuest and RequiresAuthentication) and programmatic code.
- Check if subject has a permission or not, also using the RequiresPermissions annotation and the Subject#checkPermission method.
- Check if subject has a role or not, also using the RequiresRoles annotation and Subject#checkRole
- Login and logout procedures

The only "problem" I've found is in the dependency management: If you check the [build.gradle file here](https://github.com/lucasdillmann/shiro-spring-boot-test/blob/master/build.gradle.kts#L30) you will see that I needed to explicitly include every module with the jakarta variant and exclude the regular one. This happens because of the POM file contents, which bring the regular JAR as transitive.

I don't know if this is expected or not and, if it isn't, if it should be changed or not.

lucasdillmann:
Sure. PR code is #632.

PR #574, opened 2022-12-19
[SHIRO-900] Refactor and update pom.xml

lprimak:
- Update Apache parent to 29 from 28
- Factor out jacoco properties and separate from maven surefire / failsafe argLine
- Update Groovy from Codehaus to Apache Groovy (4.x)
- Updade gmaven groovy plugin
- Update maven versions plugin
- Add compiler deprecation and unchecked lint by default
- Add JDK 11+ options for surefire and failsafe plugins

 - [x] Run `mvn clean install apache-rat:check` to make sure basic checks pass. A more thorough check will be performed on your pull request automatically.
 
### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
     <parent>
         <groupId>org.apache</groupId>
         <artifactId>apache</artifactId>
-        <version>28</version>
+        <version>29</version>
     </parent>
 
     <groupId>org.apache.shiro</groupId>
@@ -71,12 +71,13 @@
         <buildNumber>${user.name}-${maven.build.timestamp}</buildNumber>
         <project.build.outputTimestamp>2022-03-22T23:08:15Z</project.build.outputTimestamp>
         <jacoco.skip>true</jacoco.skip>
+        <surefire.version>3.0.0-M7</surefire.version>
         <!--suppress CheckTagEmptyBody -->
         <surefire.argLine></surefire.argLine>
         <!--suppress CheckTagEmptyBody -->
         <failsafe.argLine></failsafe.argLine>
         <nexus.deploy.skip>false</nexus.deploy.skip>
-
+        <jacocoAgent/>
         <!-- non-dependency-based properties: -->
         <shiro.osgi.importRange>[2, 3)</shiro.osgi.importRange>
 
@@ -118,8 +119,8 @@
         <!-- Test 3rd-party dependencies: -->
         <easymock.version>5.1.0</easymock.version>
         <mockito.version>4.8.0</mockito.version>
-        <gmaven.version>1.13.0</gmaven.version>
-        <groovy.version>3.0.13</groovy.version>
+        <gmaven.version>2.1.0</gmaven.version>
+        <groovy.version>4.0.7</groovy.version>
         <junit.version>5.9.1</junit.version>
         <junit.server.jetty.version>0.11.0</junit.server.jetty.version>
         <hibernate.version>5.6.12.Final</hibernate.version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -330,13 +333,15 @@
                             <exclude>**/spring.factories</exclude>
                             <exclude>**/spring.provides</exclude>
                             <exclude>**/*.iml</exclude>
+                            <exclude>**/nb-configuration.xml</exclude>
+                            <exclude>**/faces-config.NavData</exclude>
```
bmarwell, line 337: This could also be in the `.gitignore` file.

pom.xml
```diff
@@ -543,6 +548,12 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.8.0</version>
+                <configuration>
+                    <compilerArgs>
+                        <arg>-Xlint:deprecation</arg>
+                        <arg>-Xlint:unchecked</arg>
+                    </compilerArgs>
+                </configuration>
```
bmarwell, line 556: what about maven.compiler.showWarnings (property) or showWarnings (configuration item)? Is it set, too?

pom.xml
```diff
@@ -577,6 +588,9 @@
                                 <excludes>
                                     <exclude>**/main/**/samples/**</exclude>
                                 </excludes>
+                                <address>localhost</address>
+                                <port>${jacocoPort}</port>
+                                <propertyName>jacocoAgent</propertyName>
```
bmarwell, line 593: Why is this needed? Why was it not needed before?

pom.xml
```diff
@@ -1462,6 +1474,8 @@
                         <exclude>**/spring.factories</exclude>
                         <exclude>**/spring.provides</exclude>
                         <exclude>**/*.iml</exclude>
+                        <exclude>**/nb-configuration.xml</exclude>
+                        <exclude>**/faces-config.NavData</exclude>
```
bmarwell, line 1478: Same as above

pom.xml
```diff
@@ -1575,6 +1589,16 @@
...
             </build>
         </profile>
+        <profile>
+            <id>jdk11plus</id>
+            <activation>
+                <jdk>[11,)</jdk>
+            </activation>
+            <properties>
+                <failsafe.argLine>--add-opens java.base/java.lang=ALL-UNNAMED --add-opens java.base/java.net=ALL-UNNAMED</failsafe.argLine>
+                <surefire.argLine>-Xshare:off ${failsafe.argLine}</surefire.argLine>
```
bmarwell, line 1599: -1, `-Xshare:off` will not work on all JDKs (like non-hotspot-JDKs as OpenJ9 which comes with IBM Semeru). Either additionally check for a hotspot JDK or just leave it out. It should not be required anyway.

pom.xml
```diff
@@ -1575,6 +1589,16 @@
                 </pluginManagement>
             </build>
         </profile>
+        <profile>
+            <id>jdk11plus</id>
+            <activation>
+                <jdk>[11,)</jdk>
+            </activation>
+            <properties>
+                <failsafe.argLine>--add-opens java.base/java.lang=ALL-UNNAMED --add-opens java.base/java.net=ALL-UNNAMED</failsafe.argLine>
```
bmarwell, line 1598: As far as I can see, this should not be needed anymore. Even then, we already have this in the guice projects so it should not live here for ALL modules. Only those which need it.

bmarwell (changes requested):
Request for the profile (last 2 comments), rest is only comments/questions.

pom.xml
lprimak: Currently, ITs do not participate in jacoco coverage. This is the infra to enable this (usage coming in the Jakarta EE PR)

pom.xml
lprimak: OpenJ9 (Semuru) works perfectly with `-Xshare:off`

pom.xml
lprimak: removed

pom.xml
lprimak: done

pom.xml
lprimak: removed

pom.xml
```diff
@@ -71,12 +71,14 @@
...
         <project.build.outputTimestamp>2022-03-22T23:08:15Z</project.build.outputTimestamp>
         <jacoco.skip>true</jacoco.skip>
+        <surefire.version>3.0.0-M7</surefire.version>
         <!--suppress CheckTagEmptyBody -->
         <surefire.argLine></surefire.argLine>
         <!--suppress CheckTagEmptyBody -->
         <failsafe.argLine></failsafe.argLine>
         <nexus.deploy.skip>false</nexus.deploy.skip>
-
+        <jacocoPort>52212</jacocoPort>
```
fpapon, line 80: Is it possible to use a random port? We can have some conflicts on CI platform.

bmarwell: Usually, it is not needed to specify a port. Even with integration tests. At least when using meecrowave, as it sits in the same JVM.

lprimak: Actually, this is no longer necessary

pom.xml
```diff
@@ -1575,6 +1599,15 @@
...
         </profile>
+        <profile>
+            <id>jdk11plus</id>
+            <activation>
+                <jdk>[11,)</jdk>
+            </activation>
+            <properties>
+                <surefire.argLine>-XX:+IgnoreUnrecognizedVMOptions -Xshare:off</surefire.argLine>
+            </properties>
+        </profile>
```
bmarwell, line 1610: Small comment about why this exists would be appreciated.

pom.xml
```diff
@@ -1462,6 +1485,7 @@
                         <exclude>**/spring.factories</exclude>
                         <exclude>**/spring.provides</exclude>
                         <exclude>**/*.iml</exclude>
+                        <exclude>**/target/**</exclude>
```
bmarwell, line 1488: Only informational: This should already be included due to `.gitignore` (but doesn't hurt here either).

bmarwell (approved):
LGTM, maybe add the comment but that doesn't prevent you from merging imho.

pom.xml
lprimak: done

pom.xml
lprimak: .gitignore wasn't working for target, because the rat plugin doesn't ignore what's underneath the directories and git does.

fpapon (approved):
@lprimak good job!

PR #728, opened 2023-02-08
Restrict classes loaded in ClassResolvingObjectInputStream to org.apa…

coheigea:
A mitigation for https://github.com/apache/shiro/security/code-scanning/2
It restricts the classes that can be deserialized in ClassResolvingObjectInputStream to org.apache.shiro.* and java.util.*.


lprimak (changes requested):
I am quite sure this isn't going to work since it will fail to deserialize client application custom principals, which could be any user-defined class

coheigea:
@lprimak I could make the list of acceptable classes configurable - would that work? Or can we guarantee that malicious applications could never "force" what can be deserialized?

lprimak:
I don't think there is a security issue. The way the deserializer is used is only for encrypted data so there is no remote attack vector at all. The data is encrypted on the server with a symmetric key

PR #824, opened 2023-04-03
Refactor

vasumakwana:
### 1.	Refactoring name: Extract method.
Location: shiro\web\src\test\java\org\apache\shiro\web\servlet\SimpleCookieTest.java\Line no. 169

Explanation: The method was extracted to make the code easy to understand and the method indicates its purpose clearly.

### 2.	Refactoring name: Rename method.
Location: shiro\web\src\test\java\org\apache\shiro\web\filter\mgt\SimpleNamedFilterListTest.java\Line no. 52
Explanation: The original method name was not descriptive enough and did not follow the standard naming convention for test methods. The new name clearly indicates what the test is checking for.

Location: shiro\web\src\test\java\org\apache\shiro\web\RestoreSystemProperties.java\Line no. 48
Explanation: The new method name is more descriptive and follows the standard naming convention for method names.

### 3.	Refactoring name: Decompose conditional.
Location: shiro\samples\spring-hibernate\src\main\java\org\apache\shiro\samples\sprhib\web\SignupValidator.java\Line no. 45

Explanation: To decompose the complicated parts of the conditional into separate methods, I have create two methods:

1.	isEmailInvalid() - This method will take in the email address as a parameter and return a boolean value indicating whether the email is invalid or not based on the provided regular expression.
2.	hasInvalidEmail() - This method will take in the command object as a parameter and call the getEmail() method to retrieve the email address. It will then use the isEmailInvalid() method to check if the email is invalid or not and return a boolean value accordingly.

### 4.	Refactoring name: Extract Class
Location: shiro\web\src\main\java\org\apache\shiro\web\filter\authc\HttpMethodsExtractor.java and shiro\web\src\main\java\org\apache\shiro\web\filter\authc\HttpAuthenticationFilter.java\Line no. 205

Explanation: I have extracted the functionality for extracting HTTP methods from options into a new class called HttpMethodsExtractor. This class has a single public method extractHttpMethodsFromOptions that takes an array of options and returns a set of HTTP methods. The private method isHttpMethod is used internally by the extractHttpMethodsFromOptions method to check if an option is a valid HTTP method. In the HttpAuthenticationFilter class, I have created an instance of the HttpMethodsExtractor class and used it to extract HTTP methods from options in the 
httpMethodsFromOptions method. This makes the code more modular and easier to maintain, as the functionality for extracting HTTP methods is now encapsulated in a separate class.	

### 5.	Refactoring name: Move method
[...]

PR #831, opened 2023-04-06
Refactor 2

vasumakwana:
### 1.	Refactoring name: Extract method.
Location: shiro\web\src\test\java\org\apache\shiro\web\servlet\SimpleCookieTest.java\Line no. 169

Explanation: The method was extracted to make the code easy to understand and the method indicates its purpose clearly.

### 2.	Refactoring name: Rename method.
Location: shiro\web\src\test\java\org\apache\shiro\web\filter\mgt\SimpleNamedFilterListTest.java\Line no. 52
Explanation: The original method name was not descriptive enough and did not follow the standard naming convention for test methods. The new name clearly indicates what the test is checking for.

Location: shiro\web\src\test\java\org\apache\shiro\web\RestoreSystemProperties.java\Line no. 48
Explanation: The new method name is more descriptive and follows the standard naming convention for method names.

### 3.	Refactoring name: Decompose conditional.
Location: shiro\samples\spring-hibernate\src\main\java\org\apache\shiro\samples\sprhib\web\SignupValidator.java\Line no. 45

Explanation: To decompose the complicated parts of the conditional into separate methods, I have create two methods:

1.	isEmailInvalid() - This method will take in the email address as a parameter and return a boolean value indicating whether the email is invalid or not based on the provided regular expression.
2.	hasInvalidEmail() - This method will take in the command object as a parameter and call the getEmail() method to retrieve the email address. It will then use the isEmailInvalid() method to check if the email is invalid or not and return a boolean value accordingly.

### 4.	Refactoring name: Extract Class
Location: shiro\web\src\main\java\org\apache\shiro\web\filter\authc\HttpMethodsExtractor.java and shiro\web\src\main\java\org\apache\shiro\web\filter\authc\HttpAuthenticationFilter.java\Line no. 205

Explanation: I have extracted the functionality for extracting HTTP methods from options into a new class called HttpMethodsExtractor. This class has a single public method extractHttpMethodsFromOptions that takes an array of options and returns a set of HTTP methods. The private method isHttpMethod is used internally by the extractHttpMethodsFromOptions method to check if an option is a valid HTTP method. In the HttpAuthenticationFilter class, I have created an instance of the HttpMethodsExtractor class and used it to extract HTTP methods from options in the 
httpMethodsFromOptions method. This makes the code more modular and easier to maintain, as the functionality for extracting HTTP methods is now encapsulated in a separate class.	

### 5.	Refactoring name: Move method.
Location: web\src\main\java\org\apache\shiro\web\filter\authc\FormAuthenticationFilter.java\Line no. 52 and 184

Explanation: By moving the isLoginSubmission method to a separate class, we have made the code more modular and improved the cohesion of the classes. Additionally, any other classes that need to check if a request is a login submission can now use the LoginUtils class, rather than duplicating the logic in their own code.

### 6.	Refactoring name: Replace Conditional with Polymorphism
Location: src/main/java/org/apache/shiro/crypto/support/hashes/argon2/Argon2Hash.java/Line no. 212

Explanation: The original code has a switch statement that checks the value of a string parameter and sets the type variable to a constant based on that value. Then, it creates an instance of Argon2Parameters using the value of type and executes the generateBytes method of Argon2BytesGenerator to generate a hash. Finally, it returns a new Argon2Hash object containing the hash and other parameters.

By replacing the switch statement with polymorphism, a common interface can be defined for all possible implementations of the generateHash method, which is responsible for generating the hash. Each subclass of Argon2Algorithm can implement the generateHash method according to the corresponding value of the string parameter, thus encapsulating the algorithm-specific logic in a single class. This way, new algorithms can be easily added by adding new subclasses without modifying the existing code.

### DIFF
--- a/crypto/support/hashes/argon2/src/main/java/org/apache/shiro/crypto/support/hashes/argon2/Argon2Hash.java
+++ b/crypto/support/hashes/argon2/src/main/java/org/apache/shiro/crypto/support/hashes/argon2/Argon2Hash.java
@@ -209,17 +209,21 @@ class Argon2Hash extends AbstractCryptHash {
             int outputLengthBits
     ) {
         final int type;
+        Argon2Algorithm algorithm;
         switch (requireNonNull(algorithmName, "algorithmName")) {
             case "argon2i":
-                type = Argon2Parameters.ARGON2_i;
+                algorithm = new Argon2iAlgorithm();
+                type = algorithm.getType();
                 break;
             case "argon2d":
-                type = Argon2Parameters.ARGON2_d;
+                algorithm = new Argon2dAlgorithm();
+                type = algorithm.getType();
                 break;
             case "argon2":
                 // fall through
             case "argon2id":
-                type = Argon2Parameters.ARGON2_id;
+                algorithm = new Argon2idAlgorithm();
+                type = algorithm.getType();
                 break;
             default:
                 throw new IllegalArgumentException("Unknown argon2 algorithm: " + algorithmName);
@@ -367,5 +371,30 @@ class Argon2Hash extends AbstractCryptHash {
                 .add("memoryKiB=" + memoryKiB)
                 .add("parallelism=" + parallelism)
                 .toString();
+    }
+}
+
+abstract class Argon2Algorithm {
+    abstract int getType();
+}
+
+class Argon2iAlgorithm extends Argon2Algorithm {
+    @Override
+    int getType() {
+        return Argon2Parameters.ARGON2_i;
+    }
+}
+
+class Argon2dAlgorithm extends Argon2Algorithm {
+    @Override
+    int getType() {
+        return Argon2Parameters.ARGON2_d;
+    }
+}
+
+class Argon2idAlgorithm extends Argon2Algorithm {
+    @Override
+    int getType() {
+        return Argon2Parameters.ARGON2_id;
     }
 }
--- a/web/src/main/java/org/apache/shiro/web/filter/authc/FormAuthenticationFilter.java
+++ b/web/src/main/java/org/apache/shiro/web/filter/authc/FormAuthenticationFilter.java
@@ -74,7 +74,7 @@ class LoginUtils {
  * and {@link #setRememberMeParam(String) rememberMe} request parameters.  It then calls
  * {@link org.apache.shiro.subject.Subject#login(org.apache.shiro.authc.AuthenticationToken) Subject.login(usernamePasswordToken)},
  * effectively automatically performing a login attempt.  Note that the login attempt will only occur when the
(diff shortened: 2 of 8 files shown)


crypto/support/hashes/argon2/src/main/java/org/apache/shiro/crypto/support/hashes/argon2/Argon2Hash.java
```diff
@@ -209,17 +209,21 @@ public static Argon2Hash generate(
...
+                algorithm = new Argon2dAlgorithm();
+                type = algorithm.getType();
                 break;
             case "argon2":
                 // fall through
             case "argon2id":
-                type = Argon2Parameters.ARGON2_id;
+                algorithm = new Argon2idAlgorithm();
+                type = algorithm.getType();
                 break;
```
bmarwell, line 227: I fail to see the advantage of  this. You create a new class, but then just extract a parameter again. Yes it adds a little bit of polymorphism, but I do not see the advantage you mentioned here.
Especially, `algorithm` is not used outside the switch, so why keep it outside? Why not just write `type = newArgon2dAlgorithm().type()`. But then, why encapsulate the type anyway?

web/src/main/java/org/apache/shiro/web/filter/authc/FormAuthenticationFilter.java
```diff
@@ -30,6 +31,39 @@
 import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletRequest;
 
+/**
```
lprimak, line 34: Just like I said in #824 Please remove this extraction. It's not deserving of it's own class

web/src/test/java/org/apache/shiro/web/servlet/SimpleCookieTest.java
```diff
@@ -157,12 +157,27 @@ private static <T extends javax.servlet.http.Cookie> T eqCookie(final T in) {
         reportMatcher(new IArgumentMatcher() {
             public boolean matches(Object o) {
                 javax.servlet.http.Cookie c = (javax.servlet.http.Cookie) o;
-                return c.getName().equals(in.getName()) &&
```
lprimak, line 160: I am not sure I see the value of this either, but I can see if this is OK too, so not too opinionated on this.

crypto/support/hashes/argon2/src/main/java/org/apache/shiro/crypto/support/hashes/argon2/Argon2Hash.java
lprimak: Same as in #824 Please remove this refactor. It brings no value.

lprimak (changes requested):
Same comments as in #824 basically

PR #955, opened 2023-06-24
[GH-904] Add Spring 6-related exclusions to Shiro BOM

### LINKED ISSUE #904: Add Spring 6-related exclusions to Shiro BOM
Need to make it easier to use Spring 6/SB3
Exclusions are needed in the BOM for starters

Also, Spring 6/ SpringBoot 3 examples are needed

Suggestion:
https://github.com/apache/shiro/issues/891#issuecomment-1581192622
(untested, for now)

Continuation of #891

fpapon:
Fix #904

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -87,6 +87,42 @@
                     </exclusion>
                 </exclusions>
             </dependency>
+            <dependency>
+                <groupId>org.apache.shiro</groupId>
+                <artifactId>shiro-spring</artifactId>
+                <classifier>jakarta</classifier>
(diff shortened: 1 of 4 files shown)


lprimak (approved):

lprimak:
Should we add a test for SpringBoot?

fpapon:
> Should we add a test for SpringBoot?

I'm trying to add a sample module but I need to go deeper to exclude spring5 vs spring6 and springboot2 vs springboot3.
May be it can be another PR?

PR #1149, opened 2023-11-02
[Test] Refactored test to use utility function to get annotation instead of array indexing

wang3820:

 - [ ] Make sure there is a [GitHub issue](https://github.com/apache/shiro/issues) filed 
 - [ ] add `fixes #XXX` if merging the PR should close a related issue.
 
### Flakiness
**Nondex** was used to check and locate the flakiness in the test. The test can be reproduced using the following command:
```shell
mvn install -pl support/cdi -am -DskipTests
mvn -pl support/cdi test -Dtest=org.apache.shiro.cdi.AnnotatedTypeWrapperTest#removeAnnotations
mvn -pl support/cdi edu.illinois:nondex-maven-plugin:2.1.1:nondex -Dtest=org.apache.shiro.cdi.AnnotatedTypeWrapperTest#removeAnnotations
``` 

### Issue
The test checks that if there is no `Annotation` in the `RequiresGuest` class and there is `Annotation` in `SessionScoped` class. It retrieves these classes by calling `getDeclaredAnnotations` method and fix indexing on the array returned. This creates the problem. Here is the implementation of  `getDeclaredAnnotations` :
```java
public Annotation[] getDeclaredAnnotations() {
     return AnnotationParser.toArray(this.annotationData().declaredAnnotations);
}
// this.annotationData().declaredAnnotationsdeclaredAnnotations is a map
Map<Class<? extends Annotation>, Annotation> declaredAnnotations = ....
```
Since the array is created from a hashmap, the order of elements in the output of `getDeclaredAnnotations` is not determinstic. Therefore, the fix indexing creates flakiness.

## Fix
Since only `SessionScoped`  and `RequiresGuest` are tested, those 2 classes are specificly extracted from the array instead of using indexing.

### DIFF
--- a/support/cdi/src/test/java/org/apache/shiro/cdi/AnnotatedTypeWrapperTest.java
+++ b/support/cdi/src/test/java/org/apache/shiro/cdi/AnnotatedTypeWrapperTest.java
@@ -14,6 +14,8 @@
 package org.apache.shiro.cdi;
 
 import java.io.Serializable;
+import java.lang.annotation.Annotation;
+import java.util.Arrays;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -29,7 +31,6 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
-
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.Mock;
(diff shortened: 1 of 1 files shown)


support/cdi/src/test/java/org/apache/shiro/cdi/AnnotatedTypeWrapperTest.java
```diff
@@ -101,9 +103,23 @@ void twoAdditionalAnnotations() {
...
+            }
+        }
+        for (Annotation annotation: Annotated.class.getDeclaredAnnotations()) {
+            if (annotation.annotationType().equals(RequiresGuest.class)) {
+                s2 = Set.of(annotation);
+                break;
+            }
+        }
+        assertNotNull(s1);
+        assertNotNull(s2);
```
bdemers, line 121: I'm +1 on this, but I'll add a nit/recommendation too

I'd suggest looking up the annotation using a stream instead of the for/if/break combo.

Maybe something like this (not tested at all)

```java
Arrays.stream(SessionScopedAnnotated.class.getDeclaredAnnotations())
    .filter(a -> annotation.annotationType().equals(SessionScoped.class))
    .collect(Collectors.toSet());
```

(this could be cleaned up a bit with a `findFirst` and other tweaks)


In either case (feel free to keep the loop), you could break this logic out into a utility method so that the test logic reads well.

```java
Set<Annotation> sessionScopeAnnoations = getAnnotation(Class<?> annotatedClass, Class<?> annotation);
```

lprimak: Agreed. This is a lot more code than necessary/

wang3820: Thanks for the recommendations! Fixed.

lprimak (approved):

lprimak:
@wang3820 Can you please squash the commits? Thank you!

lprimak:
@wang3820 I have looked a little deeper into this. I believe it's a false positive actually.
The `Map` involved is actually an instance of `LinkedHashMap` so the order will never change.

Let me know if you still want me to merge this PR or just close it.
Thank you.

wang3820:
> @wang3820 I have looked a little deeper into this. I believe it's a false positive actually. The `Map` involved is actually an instance of `LinkedHashMap` so the order will never change.
> 
> Let me know if you still want me to merge this PR or just close it. Thank you.

Let me look deeper into this too. Nondex should never shuffle order of elements in a LinkedHashMap

bdemers (approved):

bdemers:
> Let me look deeper into this too. Nondex should never shuffle order of elements in a LinkedHashMap

Even so, your change makes it obvious what is going on (e.g. looking for a specific annotation class vs just grabbing one by index)

lprimak:
The other issue is that the test uses indexes in the whole code, not just this piece. Do we want to change the whole test to be consistent?

wang3820:
> > @wang3820 I have looked a little deeper into this. I believe it's a false positive actually. The `Map` involved is actually an instance of `LinkedHashMap` so the order will never change.
> > Let me know if you still want me to merge this PR or just close it. Thank you.
> 
> Let me look deeper into this too. Nondex should never shuffle order of elements in a LinkedHashMap

It appears that the test is no longer flaky. I think I was testing on a older version.

wang3820:
> The other issue is that the test uses indexes in the whole code, not just this piece. Do we want to change the whole test to be consistent?

I think it's a good idea to change the whole test. This way the test is more readable.

lprimak:
@wang3820 are you willing to take that on?

wang3820:
> @wang3820 are you willing to take that on?

Yes.

wang3820:
@bdemers @lprimak I have changed all the tests so it is consistent. 

I am rather new to this so I would really appreciate any suggestions!

lprimak (approved):

lprimak:
@wang3820 Awesome job! Thank you!
I will squash the commits and merge

bdemers:
Thanks @wang3820!

PR #1193, opened 2023-12-02
[SHIRO-820] - Fix Log4j2 logs format

ksilisk:
When using the utility, I noticed that the logs were displayed incorrectly. 
I found a similar problem in Jira, but I think the `[INFO ]` output is correct, but the space after it is a consequence of incorrect logger settings, which I corrected.
Also update log message.

JIRA issue https://issues.apache.org/jira/browse/SHIRO-820

for the change (usually before you start working on it). Trivial changes like typos do not
require a JIRA issue. Your pull request should address just this issue, without pulling in other changes.
 
### DIFF
--- a/tools/hasher/src/main/resources/log4j2.xml
+++ b/tools/hasher/src/main/resources/log4j2.xml
@@ -39,7 +39,7 @@
 
     <Appenders>
         <Console name="Console" target="SYSTEM_OUT">
-            <PatternLayout pattern="[%-5level] %msg%n"/>
+            <PatternLayout pattern="%msg%n"/>
         </Console>
     </Appenders>
     <Loggers>


bmarwell (dismissed):
Interesting. looking at this, the old hasher CLI tool just gave the hash as the output. Now, we get an additional `[INFO ]` prefix. For parsing, leaving out the prefix would be beneficial.

Then, the `-5level` is probably to align output nicely:

```
[DEBUG] message
[INFO ] hash: …
[WARN ] a warning
[ERROR] an error
```

That said, I'd prefer to rather change the pattern in `log4j.xml` to just `%msg%n` for compatibility and leave the `LOG.info` statement as-is for easier parsing. Interesting. looking at this, the old hasher CLI tool just gave the hash as the output. Now, we get an additional `[INFO ]` prefix. For parsing, leaving out the prefix would be beneficial.

Then, the `-5level` is probably to align output nicely:

```
[DEBUG] message
[INFO ] hash: …
[WARN ] a warning
[ERROR] an error
```

That said, I'd prefer to rather change the pattern in `log4j.xml` to just `%msg%n` for compatibility and leave the `LOG.info` statement as-is for easier parsing.

ksilisk:
> Interesting. looking at this, the old hasher CLI tool just gave the hash as the output. Now, we get an additional `[INFO ]` prefix. For parsing, leaving out the prefix would be beneficial.
> 
> Then, the `-5level` is probably to align output nicely:
> 
> ```
> [DEBUG] message
> [INFO ] hash: …
> [WARN ] a warning
> [ERROR] an error
> ```
> 
> That said, I'd prefer to rather change the pattern in `log4j.xml` to just `%msg%n` for compatibility and leave the `LOG.info` statement as-is for easier parsing. Interesting. looking at this, the old hasher CLI tool just gave the hash as the output. Now, we get an additional `[INFO ]` prefix. For parsing, leaving out the prefix would be beneficial.
> 
> Then, the `-5level` is probably to align output nicely:
> 
> ```
> [DEBUG] message
> [INFO ] hash: …
> [WARN ] a warning
> [ERROR] an error
> ```
> 
> That said, I'd prefer to rather change the pattern in `log4j.xml` to just `%msg%n` for compatibility and leave the `LOG.info` statement as-is for easier parsing.

Thanks for the answer @bmarwell!

No, before this the hash output looked like this
```
[INFO ] some_hash
```

In my opinion, the presence of a prefix is actually not correct, as was described in the SHIRO-820 issue.

I think that the best solution would be to display the hash in `System.out` without any prefix.
Then that will solve SHIRO-820.

In this case, there is no need to change the `log4j2.xml`
Just change `LOG.info` to `System.out.println`

If you agree with my decision, I'm ready to correct it.

lprimak:
fixed as per @bmarwell

lprimak (approved):

lprimak:
Thank you @ksilisk for your contribution!

PR #1454, opened 2024-05-03
Refactor: AssertJ best practices by @timtebeek

lprimak:
Introduce AssertJ by @timtebeek 
Supersedes #1450 and #1452

 - [ ] Make sure there is a [GitHub issue](https://github.com/apache/shiro/issues) filed 
 - [ ] Format the pull request title like `[#XXX] - Fixes bug in SessionManager`,
 - [ ] add `fixes #XXX` if merging the PR should close a related issue.
 
### DIFF
--- a/core/src/main/java/org/apache/shiro/realm/ldap/JndiLdapContextFactory.java
+++ b/core/src/main/java/org/apache/shiro/realm/ldap/JndiLdapContextFactory.java
@@ -197,7 +197,7 @@ public class JndiLdapContextFactory implements LdapContextFactory {
      *
      * @return the base JNDI environment template to use when acquiring an LDAP connection (an {@link LdapContext})
      */
-    public Map getEnvironment() {
+    public Map<String, Object> getEnvironment() {
         return this.environment;
     }
 
@@ -213,8 +213,7 @@ public class JndiLdapContextFactory implements LdapContextFactory {
      *
      * @param env the base JNDI environment template to use when acquiring LDAP connections.
      */
-    @SuppressWarnings({"unchecked"})
-    public void setEnvironment(Map env) {
+    public void setEnvironment(Map<String, Object> env) {
         this.environment = env;
     }
 
@@ -488,7 +487,7 @@ public class JndiLdapContextFactory implements LdapContextFactory {
      * @return an LdapConnection
      * @throws NamingException if a problem occurs creating the connection
      */
-    protected LdapContext createLdapContext(Hashtable env) throws NamingException {
+    protected LdapContext createLdapContext(Hashtable<String, Object> env) throws NamingException {
         return new InitialLdapContext(env, null);
     }
 
--- a/core/src/main/java/org/apache/shiro/subject/SimplePrincipalMap.java
+++ b/core/src/main/java/org/apache/shiro/subject/SimplePrincipalMap.java
@@ -157,7 +157,7 @@ public class SimplePrincipalMap implements PrincipalMap {
         if (CollectionUtils.isEmpty(this.combinedPrincipals)) {
             return Collections.emptyList();
         }
-        List<Object> list = new ArrayList<Object>(this.combinedPrincipals.size());
+        List<Object> list = new ArrayList<>(this.combinedPrincipals.size());
         list.addAll(this.combinedPrincipals.values());
         return list;
     }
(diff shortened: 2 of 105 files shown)


lprimak (commented):
Unfortunately, there are many warnings. See https://jenkins.hope.nyc.ny.us/job/docker-shiro-ci/job/main/38/analysis/

lprimak:
Hi, @timtebeek 
Unfortunately, I had to revert the AssertJ refactor. There are 66 new warnings that have resulted from the compiler.

timtebeek:
It seems 51 of those warnings are in `JndiLdapContextFactoryTest`, with cases like this one:
image

For comparison, this is what was there before, with `assertEquals` taking two objects as input:
image


What would you recommend we do about such instances?

lprimak:
And once again, proper testing reveals underlying code issues :) working on it

lprimak:
yay! It's green https://jenkins.hope.nyc.ny.us/job/docker-shiro-ci/job/assert-j-best-practices/1/

lprimak:
Please pay special attention to this commit: https://github.com/apache/shiro/pull/1454/commits/798cd1851aaa5ba2bb39ad5f4dc664dd38d9b421

I believe this is safe to put into `2.0.1` version because the API doesn't really change and just the generics are made explicit.
What do you think?

bmarwell:
> Please pay special attention to this commit: https://github.com/apache/shiro/pull/1454/commits/798cd1851aaa5ba2bb39ad5f4dc664dd38d9b421
> 
> I believe this is safe to put into `2.0.1` version because the API doesn't really change and just the generics are made explicit.
> What do you think?

Urgh. By definition a breaking change! Maybe put the 3.x label on that commit?

It's probably safe though

lprimak:
My understanding of Java generics erasure is that this change is source and binary compatible with the previous... but I could be wrong.
In former case it would be safe to merge for 2.0.1 and latter it isn't :)

bmarwell:
Binary: yes. 
Source compatible: most likely yes, but not if a user explicitly defined something else like Object,Object. Then no. 

I'd be okay with this one as it is early 2.0.x for that reason. 

@bdemers @fpapon Opinions?

lprimak:
> but not if a user explicitly defined something else like Object,Object.

Since previous code was "raw" class without any parameters, they would have to cast in that case anyway...

fpapon:
I think the problem can be this one about api change:

    public void setEnvironment(Map<String, Object> env) {

For now user can put any type in the key and we can imagine that he should use a String but we are not sure.

I understand the purpose of the best practice but it's not a good idea to update our api because of unit test framework change.

bmarwell:
> I think the problem can be this one about api change:
> 
>     public void setEnvironment(Map<String, Object> env) {
> 
> For now user can put any type in the key and we can imagine that he should use a String but we are not sure.
> 
> I understand the purpose of the best practice but it's not a good idea to update our api because of unit test framework change.

Yes, that's a real no go and needs to wait until 3.

lprimak:
Yup. As disappointing as this sounds, this one will have to wait

bmarwell:
@lprimak just modify your branch to leave that one out and merge it into 2.0.1

Then do another for 3.0.0 with that change.

bdemers:
Just catching up after some travel.

Would most of the concerns here go away if we created a custom assertion for `JndiLdapContextFactory` and `PrincipalCollection`?

That way we would at least get the migration to AssertJ in (thanks again @timtebeek).
And we can create an issue/TODO to cleanup the type <> everywhere.
Speaking of... is there an OpenRewrite recipe for that? 😄

bdemers:
(I can volunteer to write the assertion class, if that resolves the issues), it would probably read something like:
```java
Shiro.assertThat(factory).containsEnvironment("foo, "bar");
```

lprimak:
Yea, that sounds great! If there are no warnings it's all good by me :)

lprimak:
I have reverted the API change commit.

bdemers:
I started hacking a custom Assertion class up, and IMHO, it's too complex for the value it provides.

The value is essentially, working around a valid compiler warning.  The better solution for that is doing something like:
`@SuppressWarnings("unchecked")` to the relevant tests. We have a bunch of suppressions in the tests already.

My current plan is to open an issue to fix the generic type defs, and mark the effected test methods with something like:

```java
@SuppressWarning("unchecked") // TODO: Remove this warning when issue #xxxx is resolved
public void someTestMethod() {
    // ...
    assertThat(env).containsEntry(Context.SECURITY_PRINCIPAL, "foo");
    // ...
}
```

Any objections to this approach?

> [!NOTE]
> I do think custom assertion classes could be useful in other places, just in this case.

lprimak:
No onjections from me.

PR #1407, opened 2024-04-12
[#SHIRO-875] Fix creating subjects with disabled session-creation

boris-petrov:
I didn't create a GitHub issue but there is [one in Jira](https://issues.apache.org/jira/browse/SHIRO-875). This PR fixes it.

PR #1514, opened 2024-05-31
[#SHIRO-875] Fix creating subjects from a `SubjectFactory` that disables session-creation

boris-petrov:
I'm not sure if I had to remove the PR template after I've read it and made sure I followed everything in it. Perhaps you might want to note that in it.

This continues the work done in #1407. I didn't think of that case in it so this PR fixes it. I didn't open a new issue for it, sorry, I guess we could reuse the old one as it's a similar case.

cc @lprimak


lprimak (approved):

bdemers (changes requested):
This change looks good, but the test doesn't look like it's asserting that no session is being created?  Can the test be improved (or commented if i'm reading it wrong)

boris-petrov:
@bdemers thanks for the feedback! The test is pretty much the same as the one above it - it doesn't do anything with the session on its own but the `buildSubject` call fails without the fix (as it calls `save` which tries creating a session). I could add a comment, sure.

fpapon (approved):
@boris-petrov thanks for your contribution!

PR #1582, opened 2024-07-10
[#1548] - redefines PrincipalCollection interface to be intended as immutable

### LINKED ISSUE #1548: [Enhancement] Mutability of PrincipalCollection and AuthenticationInfo
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/shiro/issues?q=is%3Aissue) and found no similar issues.

### Enhancement Request

First of all, this is neither a feature request, nor is it strictly a bug report -- but the latter cannot be decided, as I will explain below. I am currently evaluating Shiro for a project and have found an architectural decision in its code that I believe is dangerous.

The culprit is PrincipalCollection. This interface does not itself specify whether it allows mutable implementations, but the existence of MutablePrincipalCollection does. Code that receives a PrincipalCollection from an unknown source must therefore treat it as "potentially mutable". PCs are passed around a lot, so "unknown source" is a common case. 

Unfortunately, receiving code doesn't do that. Received PCs are passed and stored as if they cannot be modified anymore. AbstractAuthenticator even stores them in a cache. All these places are affected by later mutations to the PC.

The code that does mutate a PC, AbstractAuthenticationStrategy (from the PAM package), will either accept an MPC from its caller and mutate that, or create its own copy in case the one from the caller is not mutable. This further obscures _which_ PC gets mutated.

The current code from Shiro does not show faulty behavior from this problem because AbstractAuthenticationStrategy is the only place where PCs get mutated, and all subclasses of that class except for FirstSuccessfulStrategy only mutate an MPC they created themselves (the "aggregate"). FirstSuccessfulStrategy uses a PC returned by a Realm as the aggregate, but then doesn't actually mutate it.

**Why is this a problem?**

I think this is a problem because it is bug-free by chance, not by design. It depends on the implementations of the interfaces used to make stronger guarantees than the interfaces do -- and these implementations don't even do that, only the combined behavior of the implementations of multiple interfaces do. 

While projects using Shiro could easily trip over this, my main concern is my own project being affected after upgrading to future Shiro versions, with its internals changing in such a way that it breaks one of those guarantees that the interfaces don't make but on which correctness of the code relies, and has to rely, on.

### Describe the solution you'd like

Possible solutions:

* My preferred solution: Make PrincipalCollection immutable by design. AFAIK, the PAM package doesn't really need it to be mutable and could instead make mutated copies. The biggest problem is code using Shiro that might make use of MPCs -- it's a breaking change.
* Make defensive copies of PCs everywhere. This is ugly, it's easy to forget a place, and it might start to affect performance.
* Establish rules about PC mutability. This is the non-invasive approach and might be a first step towards making PCs immutable.

An example for a "rule about mutability" could be: PCs should be treated as immutable, unless you created the MPC yourself and keep it to yourself -- or make a clear contract with your callers that the PC is treated as mutable. AbstractAuthenticationStrategy would have to be changed to adhere to that rule.

I am happy to contribute to the solution, but I first want to make sure that this is seen as a problem by others too, and discuss which solution approach is preferred.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

janitza-mage:
Closes #1548. Declare PrincipalCollection interface to be intended to be immutable; provide an immutable implementation including a builder class; deprecate the mutable implementation; use the immutable one in SimpleAuthenticationInfo for merging.

Background: Merging AuthenticationInfo and the contained PrincipalCollections previously lead to one of the involved PrincipalCollections be selected by undefined means and being mutated, propagating the changes to other callers that did not expect such changes, including the authentication cache.

### DIFF
--- a/core/src/main/java/org/apache/shiro/authc/SimpleAuthenticationInfo.java
+++ b/core/src/main/java/org/apache/shiro/authc/SimpleAuthenticationInfo.java
@@ -20,7 +20,7 @@ package org.apache.shiro.authc;
 
 import org.apache.shiro.lang.util.ByteSource;
 import org.apache.shiro.lang.util.SimpleByteSource;
-import org.apache.shiro.subject.MutablePrincipalCollection;
+import org.apache.shiro.subject.ImmutablePrincipalCollection;
 import org.apache.shiro.subject.PrincipalCollection;
 import org.apache.shiro.subject.SimplePrincipalCollection;
 
@@ -205,10 +205,10 @@ public class SimpleAuthenticationInfo implements MergableAuthenticationInfo, Sal
         if (this.principals == null) {
             this.principals = info.getPrincipals();
         } else {
-            if (!(this.principals instanceof MutablePrincipalCollection)) {
-                this.principals = new SimplePrincipalCollection(this.principals);
-            }
-            ((MutablePrincipalCollection) this.principals).addAll(info.getPrincipals());
+            this.principals = new ImmutablePrincipalCollection.Builder()
+                    .addPrincipals(this.principals)
+                    .addPrincipals(info.getPrincipals())
+                    .build();
         }
 
         //only mess with a salt value if we don't have one yet.  It doesn't make sense
--- a/core/src/main/java/org/apache/shiro/subject/ImmutablePrincipalCollection.java
+++ b/core/src/main/java/org/apache/shiro/subject/ImmutablePrincipalCollection.java
@@ -1,0 +1,340 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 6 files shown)


bdemers (changes requested):
I took a quick pass through, it looks great! thank you!
One nit, can you add the `@Deprecated` annotation as well (to `MutablePrincipalCollection` anywhere it's used (and set `forRemoval=true`)
(and  possibly the `@deprecated` javadoc tag to `SimplePrincipalCollection`, with a note to use `ImmutablePrincipalCollection` instead.

Thanks again!!
(I'll trigger the CI builds for this)

janitza-mage:
I have added @deprecated to SimplePrincipalCollection. That was the only place where MutablePrincipalCollection is referenced.

lprimak (approved):

lprimak:
@janitza-mage Can you please restore the template for this PR?
We would need you to click appropriate checkboxes for a contribution license acceptance.

janitza-mage:
I edited the initial comment. Remarks:
* The "fixes #XXX" is N/A because no related issue gets fixed by this
* about the milestone: I have no clue what is meant by that
* the code is apache-licensed, so I ticked the first box and left the second one unticked -- is that correct?

lprimak:
Thanks.
I fixed up the remaining issues.

janitza-mage:
Is there anything left for me to do in this PR, then? I think I can't merge it myself, right?

lprimak:
No,you are good.
We need to create 3.x development branch, make some new infra etc.
It's gonna be a while until we merge this unfortunately.

PR #2475, opened 2026-01-25
[#2421] bugfix: restored ability to match passwords from Shiro 1.x that have …

### LINKED ISSUE #2421: [Bug] Stored Shiro1 passwords can not be correctly matched when private salts are used
### Search before asking

- [x] I had searched in the [issues, **including closed issues**](https://github.com/apache/shiro/issues?q=is%3Aissue) and found no similar issues.

### Environment

Java 21, shiro-core, plain java

### Shiro version

2.0.6

### What was the actual outcome?

Failure to match password

### What was the expected outcome?

Password was matched correctly

### How to reproduce

Hi, I'm having issues migrating from Shiro 1 to 2, and before anyone suggests, reseting passwords and asking users to create new ones is not a valid suggestion for my use case, I can't seem to figure out how to actually make Shiro2 work with private salts.

Take a Shiro1 hashed password generated with the 1.x version (i.e. 1.13), take the plainText password and use the passwordsMatch(String,String) method to validate the passwords.

Example saved password (`String storedPassword`): `$shiro1$SHA-256$500000$qP2nXnatLnmrr+LQeZxnQg==$ABXgt5ZDlgCRRyaxXORR/QFrpZMW77didcJ8P7AYs/M=`
Associated plain text (`String plainText`): "admin"

Setup:
```Java
PasswordService passwordService = new DefaultPasswordService();
DefaultHashService hashService = new DefaultHashService();
hashService.setDefaultAlgorithmName("SHA-256");
passwordService.setHashService(hashService);
Shiro1CryptFormat hashFormat = new Shiro1CryptFormat();
passwordService.setHashFormat(hashFormat);

// Assert true: passwordService.passwordsMatch(plainText, storedPassword) fails
```

Our saved password was created using a private salt. And previously (i.e. 1.13) the passwordsMatch method was successful because of this the passwordsMatch(plainText, savedHash) did this:

```Java
    public boolean passwordsMatch(Object plaintext, Hash saved) {
        ByteSource plaintextBytes = createByteSource(plaintext);

        if (saved == null || saved.isEmpty()) {
            return plaintextBytes == null || plaintextBytes.isEmpty();
        } else {
            if (plaintextBytes == null || plaintextBytes.isEmpty()) {
                return false;
            }
        }

        HashRequest request = buildHashRequest(plaintextBytes, saved);

        Hash computed = this.hashService.computeHash(request);

        return constantEquals(saved.toString(), computed.toString());
    }
```

Specifically the `computeHash` call, which combines the private and public salts.

The current 2.x method flow for the `passwordsMatch(Object plainText, Hash saved)` never calls `computeHash` so the passwords never match.

So either I'm missing how Shiro 2.x handles the Shiro 1 format or the way to handle stored passwords is to override the DefaultPasswordService with support for using private salts.

Can you help?

### Debug logs

_No response_

lprimak:
…private salt

fixes #2421 

### DIFF
--- a/core/src/test/groovy/org/apache/shiro/authc/credential/DefaultPasswordServiceTest.groovy
+++ b/core/src/test/groovy/org/apache/shiro/authc/credential/DefaultPasswordServiceTest.groovy
@@ -23,6 +23,7 @@ import org.apache.shiro.crypto.hash.DefaultHashService
 import org.apache.shiro.crypto.hash.Hash
 import org.apache.shiro.crypto.hash.Sha384Hash
 import org.apache.shiro.crypto.hash.Sha512Hash
+import org.apache.shiro.crypto.hash.SimpleHashProvider
 import org.apache.shiro.crypto.hash.format.HashFormatFactory
 import org.apache.shiro.crypto.hash.format.HexFormat
 import org.apache.shiro.crypto.hash.format.Shiro1CryptFormat
@@ -31,6 +32,9 @@ import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.function.Executable
 import org.junit.jupiter.api.parallel.Isolated
 
+import java.nio.charset.StandardCharsets
+
+import static org.assertj.core.api.Assertions.assertThat
 import static org.easymock.EasyMock.*
 import static org.junit.jupiter.api.Assertions.*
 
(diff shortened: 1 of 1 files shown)


lprimak:
@bmarwell is it true that Argon and BCrypt do not use private salt? Or is private salt needs to be added to the new algorithms as well?

bmarwell:
Hi, no salt needs to be nor can be supplied by the user. See:
https://security.stackexchange.com/a/222746

The reason: argon2 and bcrypt and scrypt are NOT general purpose hashes. Hashes are meant to be fast, eg for checking data integrity. They were never meant to be used for passwords. This and adding a salt was always some kind of misuse and workaround.


General purpose hashes: meant to be fast. Adding a salt protects against rainbow tables when misused for passwords. And that's why they need so many rounds.

Password hashing algorithms (like scrypt, argon 2): meant to be memory hard or CPU hard or both, slow. Specifically for protecting passwords. Usually creating a random hash internally.

So, yeah, don't supply user hashes. For the API, just ignore them. Add to the Java doc that they will be ignored for modern functions. sha1, md5 etc are to be removed in the future for this reason.

lprimak:
Thanks!
Since salt was removed from the API, I think we are good there.

lprimak:
@bmarwell great feedback, I will make it so

steinarb:
Hm... this one might actually affect both my own usage and the JDBC realm....?

I will dig into both my own code and the changes I made to the JDBC realm (both are very old) and see if I can remember, but too sleepy tonight to think clearly.

lprimak:
Wonderful. Looking forward to your feedback in the next few days :)
No need to be sleepy

steinarb:
@lprimak A question about tests: I note no tests on this PR, is there a policy for tests in shiro?

lprimak:
@steinarb There is no policy currently, but you are right, there probably should be a test added.
The only reason I didn't consider it a must-do is because this is a 1.x backward-compatibility, 
but you are making me reconsider. Thank you

lprimak:
@steinarb I have now added a test

PR #2576, opened 2026-02-22
Configure EditorConfig for groovy,cff,yaml,yml

jbampton:
Indent size and spacing set

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -26,6 +26,10 @@ charset = utf-8
 end_of_line = lf
 insert_final_newline = true
 
-[*.rdf]
+[*.{groovy,rdf}]
 indent_size = 4
+indent_style = space
+
+[*.{cff,yaml,yml}]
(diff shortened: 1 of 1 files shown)


lprimak (changes requested):
Indents should be 4, not 2 for yaml as well to be consistent with what current code has

.editorconfig
```diff
@@ -26,6 +26,10 @@ charset = utf-8
 end_of_line = lf
 insert_final_newline = true
 
-[*.rdf]
+[*.{groovy,rdf}]
 indent_size = 4
 indent_style = space
+
+[*.{cff,yaml,yml}]
+indent_size = 2
```
lprimak: should be 4 to remain consistent

lprimak: I think that's already the default. But yes, if there are 2-spaced yaml files hanging around, they should be 4-spaces.

PR #2585, opened 2026-02-27
yamllint enable `document-start` rule checking

jbampton:


fpapon:
@jbampton can you resolve conflicts please?

Issue #2758, opened 2026-06-05
[Hardening] Deprecate RandomSessionIdGenerator due to insufficient entropy (64-bit)


Allen-wick:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/shiro/issues?q=is%3Aissue) and found no similar issues.

### Enhancement Request

The `org.apache.shiro.session.mgt.eis.RandomSessionIdGenerator` class generates session IDs using `Long.toString(SecureRandom.nextLong())`, which provides only 64 bits of entropy. This falls below the modern industry minimum of 128 bits recommended by OWASP ASVS and NIST SP 800-63B for session tokens.

While Shiro's default `JavaUuidSessionIdGenerator` (UUID v4, 122-bit entropy) is secure and used out-of-the-box, the existence of `RandomSessionIdGenerator` in the public API poses a risk of misuse by developers who might explicitly configure it for legacy or perceived performance reasons.

### Describe the solution you'd like

To improve the security posture and guide developers toward secure defaults, I propose the following hardening steps:
1. Mark `RandomSessionIdGenerator` as `@Deprecated` in the next minor release, with a Javadoc comment directing users to `JavaUuidSessionIdGenerator`.
2. Add a startup WARN log if this class is explicitly instantiated or wired via `shiro.ini` / Spring configuration.
3. Remove the class entirely in the next major release (e.g., Shiro 3.x).

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

lprimak:
Do you think that this class is useful at all? Is it worth fixing it, and not deprecating it for removal?

Allen-wick:
Thanks for looking into this so quickly and for opening the deprecation PR.
To answer your questions:
1. **Is it useful at all?** 
I don't believe it has any valid use case in modern security architectures. With OWASP ASVS and NIST SP 800-63B mandating at least 128 bits of entropy for session tokens, a 64-bit `long` falls fundamentally short. Shiro already provides `JavaUuidSessionIdGenerator` (which uses UUID v4 with 122-bit entropy) as the secure, standard default. 
2. **Is it worth fixing?** 
No, I don't think it's worth fixing. Attempting to "fix" it to meet the 128-bit threshold (e.g., by concatenating multiple `nextLong()` calls or generating a numeric string) would alter its fundamental format and data type expectations. This could silently break backward compatibility for legacy applications that explicitly rely on it returning a standard 64-bit numeric string representation.
[...]

PR #2759, opened 2026-06-06
Deprecate RandomSessionIdGenerator due to insufficient entropy

wali-reheman:
Marks RandomSessionIdGenerator as @Deprecated with a Javadoc pointing to JavaUuidSessionIdGenerator. The class only provides 64 bits of entropy via Long.toString(SecureRandom.nextLong()), below the 128-bit minimum required by OWASP ASVS and NIST SP 800-63B for session tokens.

### DIFF
--- a/core/src/main/java/org/apache/shiro/session/mgt/eis/RandomSessionIdGenerator.java
+++ b/core/src/main/java/org/apache/shiro/session/mgt/eis/RandomSessionIdGenerator.java
@@ -30,7 +30,11 @@ import java.util.Random;
  * implementation is a {@link java.security.SecureRandom SecureRandom} with the {@code SHA1PRNG} algorithm.
  *
  * @since 1.0
+ * @deprecated since 2.0. Use {@link JavaUuidSessionIdGenerator} instead. This class uses only 64 bits of entropy
+ *             which is below modern industry minimums (128 bits per OWASP ASVS and NIST SP 800-63B) and poses
+ *             a security risk if used in production.
  */
+@Deprecated
 public class RandomSessionIdGenerator implements SessionIdGenerator {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(RandomSessionIdGenerator.class);


lprimak:
Hi, Wali, thanks for your contribution.

We are not quite sure that the deprecation is the right way to go.
Please feel free to voice your opinion and discuss in the issue #2758 

Also, can you please restore the PR template, the checklist is mandatory due to copyright issues.
Also, the `@since` should be `2.2.1` instead of `2.0`
For now, let's put this PR on hold.

core/src/main/java/org/apache/shiro/session/mgt/eis/RandomSessionIdGenerator.java
```diff
@@ -30,7 +30,11 @@
  * implementation is a {@link java.security.SecureRandom SecureRandom} with the {@code SHA1PRNG} algorithm.
  *
  * @since 1.0
+ * @deprecated since 2.0. Use {@link JavaUuidSessionIdGenerator} instead. This class uses only 64 bits of entropy
```
lprimak, line 33: Can you change this to `@since 2.2.1` to make it accurate

core/src/main/java/org/apache/shiro/session/mgt/eis/RandomSessionIdGenerator.java
```diff
@@ -30,7 +30,11 @@
...
+ *             which is below modern industry minimums (128 bits per OWASP ASVS and NIST SP 800-63B) and poses
+ *             a security risk if used in production.
  */
+@Deprecated
```
lprimak, line 37: This should be annotated with `forRemoval=true` and `since=2.2.1` as well

lprimak (changes requested):
Please see the comments and these suggestions 
Thank you!

lprimak:
@wali-reheman Can you please follow-up on this? We need this pretty quick so let us know

PR #2807, opened 2026-06-22
[#2804] bugfix: fix login() once again accepts null subject

### LINKED ISSUE #2804: [Bug] DefaultSecurityManger.login(null, token) fails in 2.2.1
### Search before asking

- [x] I had searched in the [issues, **including closed issues**](https://github.com/apache/shiro/issues?q=is%3Aissue) and found no similar issues.

### Environment

We integrate Shiro into our application login, no containers are used. Jdk 21. 

### Shiro version

2.2.1

### What was the actual outcome?

In previous versions, passing a null subject into `DefaultSecurityManager.login()` would succeed. After #2704 fix (9371115) , this now fails with an NPE:

There is also no javadoc on the subject parameter, so it's not easy to tell if it was supposed to be mandatory or not. It doesn't seem like you should need a subject to login. 

```
java.lang.NullPointerException: Cannot invoke "org.apache.shiro.subject.Subject.getSession(boolean)" because "subject" is null
	at org.apache.shiro.mgt.DefaultSecurityManager.beforeSuccessfulLogin(DefaultSecurityManager.java:308)
	at org.apache.shiro.mgt.DefaultSecurityManager.login(DefaultSecurityManager.java:294)
```

### What was the expected outcome?

A new subject should be created based on the results of the login, as per previous versions. 

### How to reproduce

As above

### Debug logs

_No response_

lprimak:
This is not a typical use case, but still a minor regression.
Checks if existing subject is null, and ignores it instead of issuing an NPE.

fixes #2804

### DIFF
--- a/core/src/main/java/org/apache/shiro/mgt/DefaultSecurityManager.java
+++ b/core/src/main/java/org/apache/shiro/mgt/DefaultSecurityManager.java
@@ -305,7 +305,7 @@ public class DefaultSecurityManager extends SessionsSecurityManager {
      * @param subject Subject
      */
     protected void beforeSuccessfulLogin(Subject subject) {
-        Session session = subject.getSession(false);
+        Session session = subject != null ? subject.getSession(false) : null;
         if (session != null) {
             Map<Object, Object> attributes = new HashMap<>();
             session.getAttributeKeys().forEach(key -> attributes.put(key, session.getAttribute(key)));


core/src/main/java/org/apache/shiro/mgt/DefaultSecurityManager.java
```diff
@@ -305,7 +305,7 @@ public Subject login(Subject subject, AuthenticationToken token) throws Authenti
      * @param subject Subject
      */
     protected void beforeSuccessfulLogin(Subject subject) {
-        Session session = subject.getSession(false);
+        Session session = subject != null ? subject.getSession(false) : null;
```
steinarb, line 308: I am partial to using Optional for stuff that require null-checks
I.e. I would have done this like this (yeah, I would have used var for the local variable as well, but I'm leaving that off for now):
```
Session session = Optional.ofNullable(subject).map(s -> s.getSession(false)).orElse(null);
```
(I guess it is debatable whether this is more readable than the code from the PR, but it is the practice I'm following these days. It feels more robust. But it works better when it is integrated with the API, i.e when the API's functions returns Optional<something> instead of a nullable something)

https://2022.javazone.no/#/program/7a7de418-d373-415c-8bb9-27cd5cd5900d
(FWIW I've never seen it myself...)

Here is the reasoning behind the talk
  https://steinar.bang.priv.no/2021/12/28/how-i-learnt-to-like-optional/
  https://steinar.bang.priv.no/2022/01/18/chaining-optionals-using-flatmap-and-map/

steinarb: With a static import of Optional.session 
```
import static java.util.Optional.ofNullable;
```
it is possible to make the expression more compact
```
        Session session = ofNullable(subject).map(s -> s.getSession(false)).orElse(null);
```

lprimak: Ordinarily I would agree with you but pragmatically I think the null check is more readable in this case.

steinarb (approved):

lprimak:
Thank you Steinar!

fpapon (approved):

boris-petrov:
I don't understand - how can one login apart from `subject.login()`? Why would such a use-case be allowed at all?

lprimak:
I would agree but it's been a public API for a long time, and docs don't prohibit it, just suggest Subject.login().
There is no easy way to make it private at this point.

So IMHO best/easiest to fix it.

PR #2810, opened 2026-06-23
chore(jenkins): clean up custom workspace directories

### LINKED ISSUE #2805: [Enhancement] Clean Jenkins PR build workspace
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/shiro/issues?q=is%3Aissue) and found no similar issues.

### Enhancement Request

Clean Jenkins PR job.

Currently, a specific Jenkins job is created for each Pull Request on Jenkins. This job is never cleaned up. it leads to fill the disk space shared among all Apache projects using the Jenkins CI.

For instance with https://ci-builds.apache.org/job/Shiro/job/Shiro-all/view/change-requests/job/PR-2765/configure , we can notice that a build is never cleaned

Image

(which is what is taking much of the space)

### Describe the solution you'd like

Clean Jenkins PR check with an heuristic matching requirements. Maybe something like either when the PR is merged or after a month of PR closed.

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

lprimak:
fixes #2805

### DIFF
--- a/.jenkins.groovy
+++ b/.jenkins.groovy
@@ -202,6 +202,16 @@ Check console output at "<a href="${env.BUILD_URL}">${env.JOB_NAME} [${env.BRANC
                         // (in this case we probably don't have to do any post-build analysis)
                         cleanBeforeCheckout()
                         script {
+                            // remove MatrixCheckout directories older than 30 days to save disk space
+                            sh '''
+                                set -eu
+                                base_dir="../../../.."
+                                if [ -d "$base_dir" ]; then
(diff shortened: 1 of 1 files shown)


lprimak:
@apupier how about this?

apupier (approved):
this will save more space effectively.

it remains the case of failing/unstable builds too which is not covered. And i really understand that it can be interesting to keep the workspace in case of failure for some time.

On Apache Camel project, i think by using cleanWs(), the workspace are cleaned (most of the time, if there is not a file with a wrong user created in the workspace) only when the build is removed which i would expect to be triggered when the job is removed

PR #2836, opened 2026-07-21
[#2837] Fail closed when getPathWithinApplication() cannot normalize the request path

### LINKED ISSUE #2837: Hardening: getPathWithinApplication() returns null for un-normalizable paths, bypassing the Shiro filter chain
## Summary

`WebUtils.getPathWithinApplication()` returns `null` when the request path normalizes above root. This `null` propagates through `PathMatchingFilterChainResolver.getChain()` which cannot match any pattern (including `/**`), and `AbstractShiroFilter.getExecutionChain()` falls back to the original container `FilterChain` — bypassing **all** Shiro security filters.

## Root Cause

Commit b90f91875 (Shiro 1.5.3, CVE-2020-11989 fix) changed `getPathWithinApplication()` to:

```java
return normalize(removeSemicolon(getServletPath(request) + getPathInfo(request)));
```

`normalize()` returns `null` when the path contains `/../` at index 0 (trying to go above root). Unlike the old implementation (which always returned a valid path), this null is never handled downstream.

## Resolution

PR #2836 fixes this by failing closed: `getPathWithinApplication()` throws `IllegalStateException` when `normalize()` returns `null` or an empty path, so the request is rejected instead of being served through the unfiltered container chain. Throwing — rather than remapping the path to `/` — also avoids selecting a dedicated (often more permissive) `/` chain when a traversal request is normalized to the application root.

xaccefy:
## Problem

`WebUtils.getPathWithinApplication()` returns `null` when `normalize()` fails — e.g. for path traversal above root (`servletPath="/"` + `pathInfo="../"`). The `null` propagates through `PathMatchingFilterChainResolver.getChain()`, which matches zero patterns (including `/**`) and returns `null`, causing `AbstractShiroFilter` to fall back to the original container `FilterChain` — bypassing **all** Shiro filters, including global ones like `InvalidRequestFilter`.

Introduced in commit b90f91875 (Shiro 1.5.3, CVE-2020-11989 fix) when `normalize()` was added to `getPathWithinApplication()` without a null guard; unhandled through all subsequent versions (1.5.3 → 3.0.0).

## Fix

Fail closed: `getPathWithinApplication()` now throws `IllegalStateException` when `normalize()` returns `null` or an empty path. The exception propagates through the resolver — and through the Guice `SimpleFilterChainResolver`, which calls `WebUtils` directly — to `AbstractShiroFilter`, which fails the request closed instead of serving it through the original, unfiltered chain.

Throwing (rather than remapping the path to `/`) also avoids selecting a dedicated `/` chain when a traversal request is normalized to the application root.

## Tests

- `NullNormalizationBypassPocTest` — reproduces the bypass scenario: traversal throws in `WebUtils`, the resolver fails closed (never returns null), and normal requests still resolve
- `WebUtilsTest` — traversal-above-root paths now assert the exception
- `PathMatchingFilterChainResolverTest` — traversal request throws instead of silently matching nothing

## Checklist

fixes #2837

### DIFF
--- a/web/src/main/java/org/apache/shiro/web/filter/mgt/PathMatchingFilterChainResolver.java
+++ b/web/src/main/java/org/apache/shiro/web/filter/mgt/PathMatchingFilterChainResolver.java
@@ -150,15 +150,6 @@ public class PathMatchingFilterChainResolver implements FilterChainResolver {
             }
         }
 
-        // If no chain matched and the requestURI is null/empty, fall back to the catch-all chain if one exists.
-        // This ensures global filters (e.g. InvalidRequestFilter) always run when the request path cannot be resolved.
-        if (requestURI == null || requestURI.isEmpty()) {
-            NamedFilterList catchAllChain = filterChainManager.getChain("/**");
-            if (catchAllChain != null) {
-                return filterChainManager.proxy(originalChain, "/**");
-            }
-        }
-
         return null;
     }
 
--- a/web/src/main/java/org/apache/shiro/web/filter/mgt/PathMatchingFilterChainResolver.java
+++ b/web/src/main/java/org/apache/shiro/web/filter/mgt/PathMatchingFilterChainResolver.java
@@ -152,7 +152,7 @@ public class PathMatchingFilterChainResolver implements FilterChainResolver {
 
         // If no chain matched and the requestURI is null/empty, fall back to the catch-all chain if one exists.
         // This ensures global filters (e.g. InvalidRequestFilter) always run when the request path cannot be resolved.
-        if (requestURI == null || "".equals(requestURI)) {
+        if (requestURI == null || requestURI.isEmpty()) {
             NamedFilterList catchAllChain = filterChainManager.getChain("/**");
             if (catchAllChain != null) {
                 return filterChainManager.proxy(originalChain, "/**");
--- a/web/src/main/java/org/apache/shiro/web/filter/mgt/PathMatchingFilterChainResolver.java
+++ b/web/src/main/java/org/apache/shiro/web/filter/mgt/PathMatchingFilterChainResolver.java
@@ -150,6 +150,16 @@ public class PathMatchingFilterChainResolver implements FilterChainResolver {
             }
         }
 
+        // If no chain matched, fall back to the catch-all chain if one exists.
+        // This ensures global filters (e.g. InvalidRequestFilter) always run,
+        // even if the requestURI is null or doesn't match any pattern.
+        if (requestURI == null || "".equals(requestURI)) {
+            NamedFilterList catchAllChain = filterChainManager.getChain("/**");
+            if (catchAllChain != null) {
+                return filterChainManager.proxy(originalChain, "/**");
+            }
+        }
+
         return null;
     }
 
--- a/web/src/main/java/org/apache/shiro/web/filter/mgt/PathMatchingFilterChainResolver.java
+++ b/web/src/main/java/org/apache/shiro/web/filter/mgt/PathMatchingFilterChainResolver.java
@@ -150,9 +150,8 @@ public class PathMatchingFilterChainResolver implements FilterChainResolver {
             }
         }
 
-        // If no chain matched, fall back to the catch-all chain if one exists.
-        // This ensures global filters (e.g. InvalidRequestFilter) always run,
-        // even if the requestURI is null or doesn't match any pattern.
+        // If no chain matched and the requestURI is null/empty, fall back to the catch-all chain if one exists.
+        // This ensures global filters (e.g. InvalidRequestFilter) always run when the request path cannot be resolved.
         if (requestURI == null || "".equals(requestURI)) {
             NamedFilterList catchAllChain = filterChainManager.getChain("/**");
             if (catchAllChain != null) {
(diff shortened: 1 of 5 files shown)


web/src/main/java/org/apache/shiro/web/util/WebUtils.java
```diff
@@ -118,7 +118,11 @@ private WebUtils() {
      * @return the path within the web application
      */
     public static String getPathWithinApplication(HttpServletRequest request) {
-        return normalize(removeSemicolon(getServletPath(request) + getPathInfo(request)));
+        String path = normalize(removeSemicolon(getServletPath(request) + getPathInfo(request)));
```
lprimak: This change is very risky, and possibly introduces bypass vectors.
I think this needs to be reverted

web/src/main/java/org/apache/shiro/web/filter/mgt/PathMatchingFilterChainResolver.java
```diff
@@ -150,6 +150,15 @@ public FilterChain getChain(ServletRequest request, ServletResponse response, Fi
             }
         }
 
+        // If no chain matched and the requestURI is null/empty, fall back to the catch-all chain if one exists.
```
lprimak: This change also needs to be introduced in support/guice/src/main/java/org/apache/shiro/guice/web/SimpleFilterChainResolver.java

lprimak (changes requested):
Thanks for your contribution.

Can you please restore the issue template? It is required for copyright purposes.

See comments below as well.

lprimak:
Hey, @xaccefy 
Are you going to work on this or should consider this PR abandoned?

xaccefy:
working deeply on it

lprimak:
Any update please?

xaccefy:
@lprimak thanks for the patience! Reworked the fix: `getPathWithinApplication` now fails closed (throws `IllegalStateException` when the path cannot be normalized) instead of returning "/", so a traversal request can never fall through to the unfiltered container chain. Kept your catch-all fallback as defense in depth and added a test for it, plus a POC test reproducing the original bypass. Pushed the updated branch - ready for review.

web/src/main/java/org/apache/shiro/web/filter/mgt/PathMatchingFilterChainResolver.java
lprimak: This change should be reverted since this case will never be hit (since exception is now thrown)

web/src/main/java/org/apache/shiro/web/util/WebUtils.java
```diff
@@ -116,9 +116,18 @@ private WebUtils() {
      *
      * @param request current HTTP request
      * @return the path within the web application
+     * @throws IllegalStateException if the path cannot be normalized, e.g. when it
+     * traverses above the root ({@code "/.."}). Callers must fail closed instead of
+     * operating on a {@code null} path, which would otherwise bypass the filter chain.
      */
     public static String getPathWithinApplication(HttpServletRequest request) {
-        return normalize(removeSemicolon(getServletPath(request) + getPathInfo(request)));
+        String servletPath = getServletPath(request);
```
lprimak: Move the next 2 lines inside the exception throwing `if()` statement so no performance regressions appear

lprimak (changes requested):
Thanks for the update. Please see the review comments.
Also, the following:
- Please rename both issue and PR to something more in line with what the actual problem is now discovered to be.
- Please change the PR (and issue) description accordingly also (See Copilot review comment)
- I have restored the PR template, please fill out as necessary.

web/src/main/java/org/apache/shiro/web/util/WebUtils.java
```diff
@@ -116,9 +116,18 @@ private WebUtils() {
...
+        String pathInfo = getPathInfo(request);
+        String path = normalize(removeSemicolon(servletPath + pathInfo));
+        if (path == null) {
```
lprimak: also add `|| path.isEmpty()` here

xaccefy:
Thanks for the review! All points addressed:

- Reverted the catch-all fallback in `PathMatchingFilterChainResolver` — agreed it is unreachable now that `getPathWithinApplication()` throws; the file is back to upstream.
- Restructured `getPathWithinApplication()` per your comments: happy path unchanged, check is now `path == null || path.isEmpty()`, and the path components for the exception message are only re-read inside the throw branch.
- Re: `SimpleFilterChainResolver` (guice) — it calls `WebUtils.getPathWithinApplication()` directly, so it is covered by the same throw; no change needed there.
- Renamed issue #2837 and this PR to match the actual problem, rewrote both descriptions, and filled out the restored template.

`mvn verify` (scoped to web + its dependencies, RAT/checkstyle/javadoc included) passes. Ready for another look.

lprimak:
Thanks @xaccefy However, I still don't see the contribution license box (es) checked.

steinarb (approved):
LGTM

xaccefy:
Both license boxes are now checked. This PR is over 20 lines, so I am also filing the Apache ICLA (https://www.apache.org/licenses/icla.pdf) to clear the pending-cla label. Thanks for the review!

xaccefy:
Just to confirm: the signed ICLA (Khaserdene Baasandavaa / xaccefy) has been emailed to secretary@apache.org. Once it is on file, the pending-cla label can be cleared. Thanks!

lprimak (approved):

lprimak:
Approved. Thank you!

